PR #16, opened 2018-07-03
Stages/input cats

joezuntz:
This PR splits the shear catalog into shear and photometry catalogs, and provides a module that generates mock versions of both directly from protoDC2.


joezuntz:
Not sure we're at the point yet where formal reviews are useful, since we don't have a working baseline yet, so I will just go ahead and merge.

damonge:
Lawless Zuntz

Issue #10, opened 2018-04-24
TXSelector with a lens sample


joezuntz:
TXSelector needs to include a lens sample.  Eli is working on a python translation of the redmagic tools - this will be finished after summer.  Need something temporary in the mean time.

rmandelb:
Maybe just the BOSS CMASS and LOWZ color cuts, to make CMASS- and LOWZ-like photometric samples as a simple test case?  The would give you a red sample that goes from z~0.15 to 0.7, which is not what we ideally want but is at least simple and well-defined?

joezuntz:
Nice, thanks!  Seems to be described here:
http://www.sdss3.org/dr9/algorithms/boss_galaxy_ts.php

rmandelb:
Yes, that's correct.  It's a bright and low number density sample, quite a bit brighter than what Eli will produce with redmagic.  But it's not bad as a place holder.

rmandelb:
LOWZ selection was implemented in DESCQA by @EiffL here: https://github.com/LSSTDESC/descqa/blob/ca3025c/descqa/DeltaSigmaTest.py#L43 (lines 43-49).  It would not be hard to use similar lines of code to implement the CMASS cuts that @joezuntz linked to.  The harder part may be making sure you have the right outputs from DM (forced cmodel magnitudes, extinction-corrected) on which to impose the cuts!

joezuntz:
This is all done now - once Eli's RedMagic code is ready we can maybe move to that.

PR #19, opened 2018-08-02
Select lens sample based on BOSS color cuts

belaa:
This PR addresses Issue #10 and adds a function `select_lens()` that identifies a subset of the shear catalog as lenses if they fall within the union of LOWZ and CMASS color cuts given here: 

http://www.sdss3.org/dr9/algorithms/boss_galaxy_ts.php 

The lenses are stored as boolean True/False values in the tomography output file. 

Any suggestions or input are greatly appreciated.


joezuntz:
We chatted on the telecon about this and you were going to modify this to make the lens selection an integer (e.g. -1 = not selected, 0 = LOWZ, 1 = CMASS), then we can merge.

PR #29, opened 2019-03-01
Adding Shear-Position and Position-Position Calculations to Covariance 

ellongley:


joezuntz:
With the new sacc version this is now not relevant any more, but I did steal quite a bit from it when updating for that code.

Issue #25, opened 2019-02-08
Easy Tasks


joezuntz:
These tasks are fairly easy and a good place to start for a beginner.  Comment here or chat with @joezuntz if you're interested, or just give them a go.

- [x] Add options to twopoint stages to select which bin pairs to use
- [x] Tidy dictionary keys in covariances.py
- [x] Profiling speed of TXDiagnosticMaps with number of cores
- [ ] Investigate lens selection settings (CMASS & LOWZ) to suit this data
- [x] Write single script to generate a suite of plots from the full set of outputs

ih64:
Hi @joezuntz, just beginning to take a crack at the first bullet point. I was considering adding a 'lens_bins' and 'source_bins' options to the TXTwoPoint in the config.yaml, that expect lists. if the lists are empty, txpipe would just run all available bin pairs.  otherwise, some quick checks to make sure that the bins exist in the tomography file, and run those pairs in particular. does that sound ok?

joezuntz:
@ih64 - great, sounds good!  I think you may have to make the type a string and parse it.

joezuntz:
We now have redmagic lenses so the last missing task here is no longer needed.

PR #30, opened 2019-03-05
Ihasan/choose bins

ih64:
this branch is for #25 aimed at the first bullet point. Not yet ready to merge, but ready for comments. 

atm there is a config option for lens and source bin indices. If they are both left as empty lists, all possible bin pairs are run. if they are both lists with specified indices, then combinations of those bins are ran. Probably a neater way of handling this. Still need to add an option to handle when the user only specifies lens bins or only source bins.   

still checking if the changes made to write_output() work with sacc format, too.


txpipe/twopoint.py
```diff
@@ -55,20 +57,28 @@ def run(self):
         if self.comm:
             self.comm.Barrier()
 
-        # Get the number of bins from the
-        # tomography input file
-        nbins_source, nbins_lens  = self.read_nbins()
+        # the default is to run all bins, which is set in the config_options above
+        if self.config['source_bins'][0] == -1 and self.config['lens_bins'][0] == -1:
```
joezuntz: We might also want to deal with the case where one of the sets of bins is empty, but I see you've put that as a TODO later on so we can deal with it later.

joezuntz:
Just seen that there are some conflicts here - could you (@ih64 ) have a look at them?

txpipe/twopoint.py
ih64: sounds good, I can keep working on this development branch to include that feature

ih64:
thanks for the review! the merge conflict was a harmless one, it looks like there were other changes made to the config to add in reduce_random_size in addition to the lens and source bins list i added.

PR #31, opened 2019-04-11
Sacc2

joezuntz:
Modifies the later TXPipe stages to use the sacc2 format.

This is changed for the real and Fourier space two-point analyses, and the covariance.


joezuntz:
This is now incorporated into the extract-metacal branch

PR #35, opened 2019-06-25
Stages/extract metacal

joezuntz:
This changes the pipeline extensively in several ways:

- use the terminology and data available in the metacal catalog. 
- update to sacc2
- move parent classes for data types into the repo
- add provenance tracking tools


joezuntz:
This is too big to review, in practice, since the change to using metacal is too big to be done piecemeal anyway, plus I did another few things at the same time here, sacc2 and the provenance stuff.  That was a bad decision - I should have put them on a separate branch - but it's done now.  So I think best now is to merge and then review the operation of the provenance stuff, in particular, in a wider review in the full context.

PR #37, opened 2019-07-04
gamma_t around field centres

joezuntz:
This PR adds two stages:
- one to extract field centers from two different DBs
- one to calculate gamma_t in a single global tomographic bin, around those field centers

It needs to wait until after #36


ellongley:
Tested changes with master, no conflicts, merging.

joezuntz:
Thanks!

PR #40, opened 2019-07-19
Adding PSF maps

t-ferreira:
Adding PSF mapping.


txpipe/sysmaps.py
```diff
@@ -142,12 +143,16 @@ def run(self):
                 shear_tmp = {'mcal_g1': shear_data['true_g1'], 'mcal_g2': shear_data['true_g2']}
             else:
                 shear_tmp = {'mcal_g1': shear_data['mcal_g1'], 'mcal_g2': shear_data['mcal_g2']}
+                shear_psf_tmp = {'mcal_psf_g1': shear_data['mcal_psf_g1'], 'mcal_psf_g2': shear_data['mcal_psf_g2']}
```
joezuntz: in this line I think you need to call the dict keys just 'mcal_g1' and 'mcal_g2', because that's what the mapper is expecting (actually maybe better to modify the mapper in mapping/basic_maps.py to just look for g1 and g2 as keys and use those - cleaner)

txpipe/sysmaps.py
```diff
@@ -142,12 +143,16 @@ def run(self):
...
             shear_tmp['ra'] = phot_data['ra']
             shear_tmp['dec'] = phot_data['dec']
+            shear_psf_tmp['ra'] = phot_data['ra']       # Does it have 'ra' ?
```
joezuntz: Yes, this is right, it does need 'ra' like this.

txpipe/sysmaps.py
```diff
@@ -156,8 +161,9 @@ def run(self):
             print("Finalizing maps")
         depth_pix, depth_count, depth, depth_var = depth_mapper.finalize(self.comm)
         map_pix, ngals, g1, g2, var_g1, var_g2 = mapper.finalize(self.comm)
+        map_pix, ngals, g1, g2, var_g1, var_g2 = mapper_psf.finalize(self.comm)
```
joezuntz: need to give these new names to avoid overwriting the basic maps

txpipe/sysmaps.py
```diff
@@ -156,8 +161,9 @@ def run(self):
...
+
 
-        
         # Only the root process saves the output
         if self.rank==0:
             print("Saving maps")
```
joezuntz: (sure you know this, but for completeness - need to save the PSF map as well as the others)

PR #41, opened 2019-07-19
Stages/plots

ellongley:
Added histogram of ellipticity, and mean shear versus SNR.


joezuntz:
Hi Emily - I think this PR is the wrong way around (shouldn't it be into the plots branch?)

joezuntz:
Actually I'm completely confused now.  Are there some commits you haven't pushed from your branch yet?

PR #48, opened 2019-10-21
Added shear response calibration in diagnostics.

ellongley:


ellongley:
Very open to feedback. I'm posting the distributions of the shear response factors here as well, I'm not sure if the number of objects with zero for the response is expected. 
D3DBE122-1EF4-4895-8401-602A43F0E29C

ellongley:
Here's the distribution without any zero objects.
DCA367C0-456D-4948-8325-CB77F5BCD93F

ellongley:
And the R_total values for each bins: Interested if the off-diagonal terms are within an expected range and the differences between R11 and R22. 

R_total values: 
[[[ 0.8802616 ,  0.01803268],
        [ 0.05049371,  0.8549881 ]],

       [[ 0.7890969 , -0.01942522],
        [-0.00651014,  0.6730541 ]],

       [[ 0.8125425 , -0.0045633 ],
        [-0.03133973,  0.68719053]],

       [[ 0.8692201 ,  0.01591472],
        [-0.04687476,  0.75545317]]]

joezuntz:
I've had a look at this - your changes to twopoint.py look fine, but unfortunately the changes to the diagnostics don't quite work.

When we calculate the mean R for the diagnostics, we need a different mean R for each of the bins in the histogram (not just the tomographic bins) that gets generated, at least when the x-axis is something that shear can correlate with, like SNR.   We can't just use the global general R values.

We also have to calculate the selection bias R_S for each histogram bin, which is a bit more fiddly.

On a final note, the off-diagonal terms in the R matrix are large enough that I think we have to include them.

ellongley:
Ah! That makes much more sense, I was confused how we could calculate it from the R_total that was stored, I'll add a method to recalculate for different binnings.

joezuntz:
We also need to figure out where all the R=0 objects are coming from!

txpipe/diagnostics.py
```diff
@@ -335,19 +327,33 @@ def plot_size_shear(self):
...
-                calc1.add_data(i, data['mcal_g1'][qual_cut][w]/r_tot[0][0])
-                calc2.add_data(i, data['mcal_g2'][qual_cut][w]/r_tot[1][1])
+                w = np.where(b==i)
+                w_1p = np.where(b_1p==i)
+                w_2p = np.where(b_2p==i)
+                w_1m = np.where(b_1m==i)
+                w_2m = np.where(b_2m==i)
+                # Calculate the shear response 
+                S = calculate_selection_bias(data['mcal_g1'][qual_cut], data['mcal_g2'][qual_cut], w_1p, w_2p,w_1m, w_2m, delta_gamma)
+                R = calculate_multiplicative_bias(data['mcal_g1_1p'],data['mcal_g1_2p'],data['mcal_g1_1m'],data['mcal_g1_2m'],
```
joezuntz: Does the qual_cut also need to be on these lines?

txpipe/diagnostics.py
```diff
@@ -1,7 +1,7 @@
 from .base_stage import PipelineStage
 from .data_types import Directory, HDFFile, PNGFile, TomographyCatalog
 from .utils.stats import ParallelStatsCalculator
-from .utils.metacal import apply_response
+from .utils.metacal import calculate_selection_bias, calculate_multiplicative_bias, apply_metacal_response
```
joezuntz: Thanks for putting these in a wider library, they'll be v useful for other stuff elsewhere I think.

txpipe/diagnostics.py
```diff
@@ -335,19 +327,33 @@ def plot_size_shear(self):
...
+            b_2p = np.digitize(data['mcal_T_2p'][qual_cut], T_edges) - 1
+            b_1m = np.digitize(data['mcal_T_1m'][qual_cut], T_edges) - 1 
+            b_2m = np.digitize(data['mcal_T_2m'][qual_cut], T_edges) - 1
 
             for i in range(size):
-                w = np.where(b1==i)
-                # Do more things here to establish
-                calc1.add_data(i, data['mcal_g1'][qual_cut][w]/r_tot[0][0])
-                calc2.add_data(i, data['mcal_g2'][qual_cut][w]/r_tot[1][1])
+                w = np.where(b==i)
```
joezuntz: Need to make an equivalent change in the S2N cut also.

txpipe/utils/metacal.py
```diff
@@ -13,7 +13,52 @@ def metacal_band_variants(bands, *names):
...
+    
+    # Also save the selection biases as a matrix.
+    S[0,0] = S_11
+    S[0,1] = S_12
+    S[1,0] = S_21
+    S[1,1] = S_22
+    
+    return S
+
+def calculate_multiplicative_bias(g1_1p,g1_2p,g1_1m,g1_2m,g2_1p,g2_2p,g2_1m,g2_2m,sel_00,delta_gamma):
```
joezuntz: They're both "multiplicative" biases, but I can't think of a better name for this.  "Response bias"?

Also I'm wondering if there's something we could do to simplify passing around these long lists of arguments.  Let's think about that later though.

joezuntz:
Ack, sorry - just had a realization about this that I should have thought of before - the selection response S needs to be calculated over all the chunks of data, not just within a chunk.  i.e. the applied at the end of the `While True` loop.  We need to build up the total R from all the chunks.
What you have now should be more accurate than what's there already, as long as the chunks are large enough (which we control), so let's merge this for now and then figure out the best way to fix.

txpipe/utils/metacal.py
ellongley: In general this stage has some large arguments being passed around and long lines of similar actions, I can look into cleaning it up for the the future.

ellongley:
Thanks for that catch, I've made an issue so we know we need to fix this.

PR #52, opened 2019-11-18
Updated welcome to TXPipe notebook.

ellongley:


joezuntz:
looks like there is currently an error in the TXCov reading stage

PR #58, opened 2020-01-11
Add option to use input pz instead of running MLZ

chihway:
In selector.py add the option to use an input redshift point-estimate for each galaxy to do the binning. This is determined by the parameter `input_pz`  set in the configuration file. `input_pz: False` will give the original pipeline. 

- requires a pre-existing photoz_stack.hdf5 file 
- skip "PZPDFMLZ" and "TXPhotozStack" stages if `input_pz: True`

Example for DES Y1 is described [here](https://github.com/LSSTDESC/txpipe-reanalysis/issues/6). Other experiments will do this differently since they do not have metacal photo-z's. Yaml and config files [here](https://github.com/LSSTDESC/txpipe-reanalysis/blob/master/desy1/desy1_input_pz.yml) and [here](https://github.com/LSSTDESC/txpipe-reanalysis/blob/master/desy1/config_input_pz.yml).

### DIFF
--- a/txpipe/selector.py
+++ b/txpipe/selector.py
@@ -37,6 +37,7 @@ class TXSelector(PipelineStage):
     ]
 
     config_options = {
+        'input_pz': False,
         'bands': 'riz', # bands from metacal to use
         'T_cut':float,
         's2n_cut':float,
@@ -82,9 +83,9 @@ class TXSelector(PipelineStage):
         bands = self.config['bands']
         chunk_rows = self.config['chunk_rows']
 
-
-        # Build a classifier used to put objects into tomographic bins
-        classifier, features = self.build_tomographic_classifier()        
+        if self.config['input_pz'] == False:
+            # Build a classifier used to put objects into tomographic bins
+            classifier, features = self.build_tomographic_classifier()        
 
         # Colums we need from the photometry data.
         # We use the photometry data to select the lenses.
@@ -95,6 +96,8 @@ class TXSelector(PipelineStage):
         shear_cols = ['mcal_flags', 'mcal_psf_T_mean']
         shear_cols += metacal_band_variants(bands, 'mcal_mag', 'mcal_mag_err')
         shear_cols += metacal_variants('mcal_T', 'mcal_s2n', 'mcal_g1', 'mcal_g2')
+        if self.config['input_pz'] == True:
+            shear_cols += ['mean_z']
 
         # Input data.  These are iterators - they lazily load chunks
         # of the data one by one later when we do the for loop.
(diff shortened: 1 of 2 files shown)


txpipe/selector.py
```diff
@@ -81,10 +82,11 @@ def run(self):
         # various config options
         bands = self.config['bands']
         chunk_rows = self.config['chunk_rows']
+        #print(self.config['input_pz'])
```
joezuntz: This can be deleted

txpipe/selector.py
```diff
@@ -81,10 +82,11 @@ def run(self):
...
 
-
-        # Build a classifier used to put objects into tomographic bins
-        classifier, features = self.build_tomographic_classifier()        
+        if self.config['input_pz'] == False:
```
joezuntz: `if x == True` in python is better expressed as just `if x`.  Similarly for `if x == False` it's better to just do `if not x`

txpipe/selector.py
```diff
@@ -95,6 +97,13 @@ def run(self):
         shear_cols = ['mcal_flags', 'mcal_psf_T_mean']
         shear_cols += metacal_band_variants(bands, 'mcal_mag', 'mcal_mag_err')
         shear_cols += metacal_variants('mcal_T', 'mcal_s2n', 'mcal_g1', 'mcal_g2')
+        if self.config['input_pz'] == True:
```
joezuntz: See note above

txpipe/selector.py
```diff
@@ -117,9 +125,13 @@ def run(self):
...
             print(f"Process {self.rank} running selection for rows {start:,}-{end:,}")
-
-            # Select most likely tomographic source bin
-            pz_data = self.apply_classifier(classifier, features, shear_data)
+            
+            if self.config['input_pz'] == False:
+              # Select most likely tomographic source bin
+              pz_data = self.apply_classifier(classifier, features, shear_data)
+            
+            if self.config['input_pz'] == True:
```
joezuntz: Neater to do this with an `else`

txpipe/selector.py
```diff
@@ -244,6 +256,22 @@ def apply_classifier(self, classifier, features, shear_data):
...
 
+    def apply_simple_redshift_cut(self, shear_data):
+        variants = ['', '_1p', '_2p', '_1m', '_2m']
+
+        pz_data = {}
+
+        for v in variants:
+            zz = shear_data[f'mean_z{v}']
+
+            pz_data_v = np.zeros(len(zz)) -1
```
joezuntz: think you need a `dtype=int` here

joezuntz:
Thanks @chihway !

PR #63, opened 2020-01-24
Muir et al type blinding

slosar:
Progress so far:
 * Basic functionality to TwoPoint stage
 * Config plumbing and cosmology works 
 * Some work towards getting firecrown to make predictions
 * Comments welcome.


joezuntz:
I don't think the code will be happy using the same name for the input and output files. 
You probably need to call the output one "twopoint_data_blind' or something.

slosar:
This is now a working, e.g. running version of muir type blinding. It uses firecrown to make predictions. It requires the `parser_mods` branch of Firecrown. It hasn't really been validated, but I inspected a few numbers and they seem to be changing my sensible amounts. 
The interface to firecrown is somewhat klunky, but not terrible. 
For me this was really an exercise in how these things fit together.

txpipe/blinding.py
```diff
@@ -0,0 +1,226 @@
+###
+#
+# TO BE DEPRECATED VERY SOON
+# Leaving here to be able to see what is going on
```
joezuntz: Delete this if it's not used now.

joezuntz:
will add some more notes on the code itself though.

txpipe/blinding.py
slosar: Bah, I thought i've deleted that. Should be gone now.

slosar:
The reason why I went against TXBlinding were:
 * You cannot make blinding default, so every person forgetting TXBlinding will start analyzing unblinded data. To me this just sounds a too easy barrier to breach by mistake
 * In my understanding each stage has an input and an output. To have TXBlinding as a separate stage, you need to have unblinded files lying around, which again are easy to breach. You could name them very prominently though!

Basically, if you see a way of enforcing TXBlinding stage unless you really want to disable it, then it might actually be not too bad.

txpipe/twopoint.py
```diff
@@ -202,9 +221,118 @@ def _read_nbin_from_config(self):
...
+                                   sources: ['{tracer1}', '{tracer2}']
+                                   sacc_data_type: {dtype}
+                   """.format(fcname=fcname, dtype=dtype,tracer1=tracer1,tracer2=tracer2)
+        ## now try to get predictions
+        pred={}
+        for name,pars in [('fid',fid_params),('ofs',offset_params)]:
+            print ("Calling firecrown : %s"%(name))
+
+            config, data = firecrown.parse(io.StringIO(fc_config),
+               settings={'two_point':{'sacc_data':sacc},'parameters':pars})
```
beckermr: So @slosar simply make the edits you'd like to config and data here in the code. And then feed those to firecrown. This will keep the code I don't like out of firecrown.

slosar: Ok, I can pass cosmological parameters as strings rather than floats (it is ugly in my view, but ok), but I cannot pass a sacc object as a string. The only way to do it is dump it to disc, which is to start with very inefficient at nersc (takes minutes to dump a 2Mb sacc file), but important it is just wrong to be interacting with a library via objects on disk.

beckermr: Right so sacc object is def on the firecrown side. I will push a PR.

joezuntz:
On the call we agreed to move this into TXBlinding

slosar:
Now with a separate stage and with new firecrown interface (which does look much nicer!).
One side effect is that after TXBlinding deletes the unblided file, ceci wants to regenerate it next time. Not sure what to do about that. @joezuntz

slosar:
Now it doesn't delete the unblinded file, but instead it rewrites it with zero size one.

txpipe/blinding.py
```diff
@@ -0,0 +1,157 @@
...
+                }
+        }
+
+
+        for k,v in bz.items():
+            fc_config ['parameters']['bias_%s'%k] = v
+            
+        srclist=[]
+        lenslist=[]
+        for key,tracer in sacc.tracers.items():
```
joezuntz, line 112: What you're trying to do for this whole chunk is not clear on first reading - can you add a lot more comments explaining what you're doing and how?

joezuntz:
thought of some more things - hold fire.

txpipe/blinding.py
```diff
@@ -0,0 +1,157 @@
...
+import warnings
+
+class TXBlinding(PipelineStage):
+    """
+    Blinding the data vectors.
+
+    """
+    name='TXBlinding'
+    inputs = [
+        ('twopoint_data_UNBLINDED_RABID', SACCFile),
```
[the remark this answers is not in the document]
joezuntz: Actually - we do need a suffix here, because otherwise the unblind file will be used for later stages.
Argh.  Actually, I think I need to think about this.

Also in the master branch we've change `twopoint_data` -> `twopoint_data_real`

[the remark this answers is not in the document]
joezuntz: Okay, think the answer here is to change the suffix on the original file to `_raw` and then also add a null blinding stage that just copies `twopoint_data_real_raw` -> `twopoint_data_real` (and overwrites the raw file with an empty one, as is done now).

Then `raw` doesn't indicate that it definitely will be blinded later, but is neutral on the question.

ellongley:
Hi all, @chihway and I are volunteering to help on this branch as we need to learn to run the blinding stage for the reanalysis work. Here is an attempt at addressing the latest comments - feedback appreciated! :)

slosar:
hi @ellongley  -- great to see progress. I really honestly had intention to finish this and will eventually, but if you beat me to it, even better! feel free to hassle me!

txpipe/blinding.py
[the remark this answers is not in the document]
ellongley: Are these now the correct suffixes? Is there a way to copy the file using the TXPipe object or should it just be with shutil e.g.?

txpipe/blinding.py
```diff
@@ -0,0 +1,157 @@
...
+        sys.stdout.flush()
+
+        if self.rank==0:
+            unblinded_fname = self.get_input('twopoint_data_UNBLINDED_RABID')
+            sacc = sacc.Sacc.load_fits(unblinded_fname)
+            blinded_sacc = self.blind_muir(sacc)
+            print ("Writing a very small sacc file [on NERSC: 1200 baud teetu-teetu-shhhhh]")
+            blinded_sacc.save_fits(self.get_output('twopoint_data'), overwrite=True)
+            if self.config['delete_unblinded']:
+                print ("Replacing %s with empty..."%unblinded_fname)
```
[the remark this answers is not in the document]
ellongley: I _think_ I caught all the prints.

txpipe/blinding.py
```diff
@@ -0,0 +1,157 @@
...
+        
+         - Load two point SACC file
+         - Blinding it 
+         - Output blinded data
+         - Optionally deletete unblinded data
+        """
+        import sacc, sys, os
+        sys.stdout.flush()
+
+        if self.rank==0:
```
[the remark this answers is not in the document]
ellongley: I deleted the test, but if there is something to add to mark to just run in serial that can be done. :)

joezuntz:
One last request - could you add a null blinding step to the examples?

PR #64, opened 2020-01-24
Lensfit calibrator

ellongley:
Added option to apply nonmetacal or metacal shear calibration. With a placeholder lens fit calibration for now.


txpipe/diagnostics.py
```diff
@@ -372,7 +373,12 @@ def plot_g_histogram(self):
                 R = calculate_shear_response(data['mcal_g1_1p'][qual_cut],data['mcal_g1_2p'][qual_cut],data['mcal_g1_1m'][qual_cut],data['mcal_g1_2m'][qual_cut],
                                                   data['mcal_g2_1p'][qual_cut],data['mcal_g2_2p'][qual_cut],data['mcal_g2_1m'][qual_cut],data['mcal_g2_2m'][qual_cut],delta_gamma)
                 
-                g1, g2 = apply_metacal_response(R, S, data['mcal_g1'][qual_cut][w1], data['mcal_g2'][qual_cut][w1])
+                if self.config['shear_catalog_type']=='metacal':
+                    g1, g2 = apply_metacal_response(R, S, data['mcal_g1'][qual_cut][w1], data['mcal_g2'][qual_cut][w1])
+                elif self.config['shear_catalog_type']=='nonmetacal':
```
joezuntz: Since it's currently calling `apply_lensfit_calibration` could you change the param name to just `lensfit`?

txpipe/diagnostics.py
```diff
@@ -30,7 +30,8 @@ class TXDiagnosticPlots(PipelineStage):
 
     config_options = {
         'chunk_rows': 100000,
-        'delta_gamma': 0.02
+        'delta_gamma': 0.02,
+        'shear_catalog_type': 'nonmetacal'
```
joezuntz: Can you leave the default as `metacal` as before?

txpipe/twopoint.py
```diff
@@ -45,7 +45,8 @@ class TXTwoPoint(PipelineStage):
         'reduce_randoms_size':1.0,
         'do_shear_shear': True,
         'do_shear_pos': True,
-        'do_pos_pos': True
+        'do_pos_pos': True,
+        'shear_catalog_type': 'nonmetacal'
```
joezuntz: Also leave default as metacal here

txpipe/twopoint.py
```diff
@@ -322,7 +323,10 @@ def get_m(self, data, i):
         mask = (data['source_bin'] == i)
         
         # We use S=0 here because we have already included it in R_total
-        g1, g2 = apply_metacal_response(data['R_total'][i], 0.0, data['mcal_g1'][mask],data['mcal_g2'][mask])
+        if self.config['shear_catalog_type']=='metacal':
+            g1, g2 = apply_metacal_response(data['R_total'][i], 0.0, data['mcal_g1'][mask],data['mcal_g2'][mask])
+        elif self.config['shear_catalog_type']=='nonmetacal':
+            g1, g2 = apply_lensfit_calibration(data['mcal_g1'][mask],data['mcal_g2'][mask])
```
joezuntz: Can you add an else here checking for unknown calibrators / wrong choice of the parameter.

txpipe/utils/calibration_tools.py
```diff
@@ -67,6 +67,19 @@ def apply_metacal_response(R, S, g1, g2):
...
 
+def apply_lensfit_calibration(g1, g2, weight=None,m=None):
+    #This function needs to be checked!
+    
+    if weight==None:
+        weight = np.array([1])
+    if m==None:
+        m = np.array([1])
+
+    return g1, g2
```
joezuntz: Did you mean to delete all the stuff that was here?

joezuntz: It definitely looks wrong now :-)

MariaElidaiana: Hi Joe. Yes, it was wrong. Emily told me she committed this version by accident and I thought she was correcting me. Anyway, the old version that I suggested it was not right either. But now, I checked some emails exchange we had with Catherine and I'm adding a new version that it seems a bit more right. It is based on this suggestion that Catherine gave to us: _"On m-corrections per galaxy - these do exist but they are pretty much nonsense....For TXPipe, I would suggest you simply attach a responsivity of 1 per galaxy, and then the actual m-correction per tomographic bin folds through as a non-zero mean m nuisance parameter prior per bin with the width of 0.02 (kannawadi et al 2019 has the values to use for the peak of the prior). We have 100% correlation between the m-errors in each bin - can TXPipe/FireCrown deal with correlated nuisance parameters?"_

So, my interpretation of this was: don't apply the m-correction (a.k.a the 1+K term) and just pass to treeCorr the lensfit weight. Although, I let the possibility to pass the individual m.

txpipe/selector.py
```diff
@@ -298,10 +345,16 @@ def calculate_tomography(self, pz_data, shear_data, calibrators):
...
-        R[:,0,1] = (data['mcal_g1_2p'] - data['mcal_g1_2m']) / delta_gamma
-        R[:,1,0] = (data['mcal_g2_1p'] - data['mcal_g2_1m']) / delta_gamma
-        R[:,1,1] = (data['mcal_g2_2p'] - data['mcal_g2_2m']) / delta_gamma
+        if self.config['shear_type']=='metacal':
+            R[:,0,0] = (data['mcal_g1_1p'] - data['mcal_g1_1m']) / delta_gamma
+            R[:,0,1] = (data['mcal_g1_2p'] - data['mcal_g1_2m']) / delta_gamma
+            R[:,1,0] = (data['mcal_g2_1p'] - data['mcal_g2_1m']) / delta_gamma
+            R[:,1,1] = (data['mcal_g2_2p'] - data['mcal_g2_2m']) / delta_gamma
+        else:
+            R[:,0,0] = 1
```
joezuntz, line 354: Some of these should be 0

txpipe/data_types/types.py
```diff
@@ -12,16 +12,36 @@ def metacalibration_names(names):
...
+            raise ValueError(f"Unable to read shear catalog")
+        shear_catalog_type = info.get('catalog_type')
+        return shear_catalog_type
 
-    # Add methods for handling here ...
+    def validate(self):
+        
+        shear_catalog_type = self.read_catalog_info()
+        
+        if shear_catalog_type=='metacal':
```
joezuntz, line 34: The required datasets here are not being used.  I think you can just delete - the validation stuff has never been as useful as I'd thought.

txpipe/selector.py
```diff
@@ -485,13 +538,19 @@ def select(self, data, bin_index):
         s2n_cut = self.config['s2n_cut']
         T_cut = self.config['T_cut']
         verbose = self.config['verbose']
-
-        s2n = data['mcal_s2n']
-        T = data['mcal_T']
+        
+        if self.config['shear_type']=='metacal':
```
joezuntz, line 542: To avoid spiralling complexity in the downstream stages I think we need to rename the metacal columns after loading them, to remove the mcal prefix and remove lots of the `if type === metacal` things.

I know that's a pain - sorry.

ellongley: I tried to take out any of the if type == metacal pertaining to the column names themselves, and left the column names without the mcal prefix - is there a section in the inputs that needs to be added to rename the columns?

PR #67, opened 2020-02-08
Jackknife

jpratmarti:
@empEvil and I have added the option to obtain jackknife covariances, using TreeCorr's new jackknife implementation, which is currently under `jk2` branch in TreeCorr repository. A docker image to test this code with this TreeCorr's branch has been created and can be loaded doing:

`shifter --image jpratmarti/txpipe-jk2 bash`

We have used that to test the code is running without errors. The main changes are the following:

- Added two more options in config_options dictionary in each of the classes inside twopoint.py:
```
        'var_methods': 'jackknife',
        'npatch': 50
```
`var_methods` can be set to 'shot' if the user prefers to run without jackknife. In principle this can be extended to all the different variance methods which are implemented now in TreeCorr, such as bootstrap, etc. 

- We have changed how results from the Treecorr calls interact with the code. This is because to calculate the combined covariance we need the treecorr objects. This means the following changes:

   -- The 3 calculate correlation functions, now only returns a object in the form of the treecorr correlation object.
   -- In the write output part of the code, it now checks what type of correlation the result it is, then it adds the treecorr object to the appropriate list of objects (gg, ng, nn) for calculating the covariance. Then it reads out the results as before and adds that to the data-point.
   -- Finally it combines the 3 lists of correlations, and then calls treecorr to calculate the combined covariance.

Things to consider in the future:

- Better way to call jackknife, and make it better to ensure that the patches are the same. This is a potential issue already.
- The 3 lists of objects are currently sorted by type, but they might not be sorted correctly by bin combinations.
- There might be an issue with that it will calculate the cov xi+ xi- for each object and then add that to combined cov and then go to the next object, which means it is not properly sorting by type, will ask Mike about this.

### DIFF
--- a/examples/laptop.yml
+++ b/examples/laptop.yml
@@ -40,6 +40,7 @@ inputs:
     calibration_table: data/example/inputs/sample_cosmodc2_w10year_errors.dat
     exposures: data/example/inputs/exposures.hdf5
     star_catalog: data/example/inputs/star_catalog.hdf5
+    patch_centers: data/example/inputs/jackcenters.hdf5
     # This file comes with the code
     fiducial_cosmology: data/fiducial_cosmology.yml
 
@@ -50,5 +51,3 @@ resume: True
 log_dir: data/example/logs
 # where to put an overall parsl pipeline log
 pipeline_log: data/example/log.txt
-
-
--- a/examples/laptop.yml
+++ b/examples/laptop.yml
@@ -5,6 +5,7 @@ stages:
     - name: TXPhotozStack
     - name: TXDiagnosticMaps
     - name: TXRandomCat
+    - name: TXJackknifecenters
     - name: TXTwoPoint
     - name: TXDiagnosticPlots
     - name: TXGammaTFieldCenters
--- a/examples/laptop.yml
+++ b/examples/laptop.yml
@@ -40,7 +40,7 @@ inputs:
     calibration_table: data/example/inputs/sample_cosmodc2_w10year_errors.dat
     exposures: data/example/inputs/exposures.hdf5
     star_catalog: data/example/inputs/star_catalog.hdf5
-    patch_centers: data/example/inputs/jackcenters.hdf5
+    #patch_centers: data/example/inputs/jackcenters.hdf5
     # This file comes with the code
     fiducial_cosmology: data/fiducial_cosmology.yml
 
--- a/examples/laptop.yml
+++ b/examples/laptop.yml
@@ -5,7 +5,7 @@ stages:
     - name: TXPhotozStack
     - name: TXDiagnosticMaps
     - name: TXRandomCat
-    - name: TXJackknifecenters
+    - name: TXJackknifeCenters
     - name: TXTwoPoint
     - name: TXDiagnosticPlots
     - name: TXGammaTFieldCenters
@@ -42,7 +42,6 @@ inputs:
     calibration_table: data/example/inputs/sample_cosmodc2_w10year_errors.dat
     exposures: data/example/inputs/exposures.hdf5
     star_catalog: data/example/inputs/star_catalog.hdf5
-    #patch_centers: data/example/inputs/jackcenters.hdf5
     # This file comes with the code
     fiducial_cosmology: data/fiducial_cosmology.yml
 
--- a/examples/laptop.yml
+++ b/examples/laptop.yml
@@ -11,6 +11,7 @@ stages:
     - name: TXGammaTBrightStars
     - name: TXGammaTDimStars
     - name: TXRoweStatistics
+    - name: TXTwoPointPlots
 
 # Where to put outputs
 output_dir: data/example/outputs
@@ -46,7 +47,7 @@ inputs:
 
 # if supported by the launcher, restart the pipeline where it left off
 # if interrupted
-resume: True
+resume: False
 # where to put output logs for individual stages
 log_dir: data/example/logs
 # where to put an overall parsl pipeline log
--- a/examples/laptop.yml
+++ b/examples/laptop.yml
@@ -47,7 +47,7 @@ inputs:
 
 # if supported by the launcher, restart the pipeline where it left off
 # if interrupted
-resume: False
+resume: True
 # where to put output logs for individual stages
 log_dir: data/example/logs
 # where to put an overall parsl pipeline log
--- a/txpipe/sysmaps.py
+++ b/txpipe/sysmaps.py
@@ -175,7 +175,6 @@ class TXDiagnosticMaps(PipelineStage):
         flag_pixs, flag_maps = flag_mapper.finalize(self.comm)
         import treecorr
         cat = treecorr.Catalog(ra = ra, dec = dec, ra_units='degree', dec_units = 'degree', npatch=self.config['npatch'])
-        cat.patches
         cat.write_patch_centers(self.get_output('patch_centers'))
 
         # Only the root process saves the output
--- a/txpipe/sysmaps.py
+++ b/txpipe/sysmaps.py
@@ -30,6 +30,7 @@ class TXDiagnosticMaps(PipelineStage):
     outputs = [
         ('diagnostic_maps', DiagnosticMaps),
         ('tracer_metdata', HDFFile),
+        ('patch_centers', HDFFile),
     ]
 
     # Configuration information for this stage
@@ -126,6 +127,10 @@ class TXDiagnosticMaps(PipelineStage):
         m_it = self.iterate_hdf('tomography_catalog','multiplicative_bias', m_cols, chunk_rows)
         m_it = (d[2] for d in m_it)
 
+        #Naive attempt at generating the Jackknife centers!
+        ra =[]
+        dec = []
+
         # Now, we actually start loading the data in.
         # This thing below just loops through all the files at once
         iterator = zip(shear_it, phot_it, bin_it, m_it)
@@ -152,7 +157,8 @@ class TXDiagnosticMaps(PipelineStage):
             shear_tmp['dec'] = phot_data['dec']
             shear_psf_tmp['ra'] = phot_data['ra']       # Does it have 'ra' ?
             shear_psf_tmp['dec'] = phot_data['dec']     # Does it have 'dec' ?
-
+            ra.extend(phot_data['ra'])
+            dec.extend(phot_data['dec'])
             # And add these data chunks to our maps
             depth_mapper.add_data(depth_data)
             mapper.add_data(shear_tmp, bin_data, m_data)
@@ -167,6 +173,10 @@ class TXDiagnosticMaps(PipelineStage):
         map_pix, ngals, g1, g2, var_g1, var_g2 = mapper.finalize(self.comm)
         map_pix_psf, ngals_psf, g1_psf, g2_psf, var_g1_psf, var_g2_psf = mapper_psf.finalize(self.comm)
         flag_pixs, flag_maps = flag_mapper.finalize(self.comm)
+        import treecorr
+        cat = treecorr.Catalog(ra = ra, dec = dec, ra_units='degree', dec_units = 'degree', npatch=self.config['npatch'])
+        cat.patches
+        cat.write_patch_centers(self.get_output('patch_centers'))
 
         # Only the root process saves the output
         if self.rank==0:
@@ -358,4 +368,3 @@ class TXMapPlots(PipelineStage):
 
 if __name__ == '__main__':
     PipelineStage.main()
-
--- a/txpipe/sysmaps.py
+++ b/txpipe/sysmaps.py
@@ -126,10 +126,6 @@ class TXDiagnosticMaps(PipelineStage):
         m_it = self.iterate_hdf('tomography_catalog','multiplicative_bias', m_cols, chunk_rows)
         m_it = (d[2] for d in m_it)
 
-        #Naive attempt at generating the Jackknife centers!
-        ra =[]
-        dec = []
-
         # Now, we actually start loading the data in.
         # This thing below just loops through all the files at once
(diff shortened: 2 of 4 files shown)


chihway:
Question for @joezuntz , currently the code uses
`S.add_covariance(cov)` to add the JK covariance to sacc, but should we modify so that there's possibilities for storing multiple covariance?

joezuntz:
Just looked at this - I think you have your covariance potentially in the wrong order.  You give treecorr thing organized by ggs then ngs then nns, but that's not necessarily the order you do S.add_data_point.  Is there a reason you need separate ggs, ngs, and nns?

joezuntz:
Also, does this work under MPI?  `collect_results` collects together the Measurement objects from across the processors, which would definitely be very very memory heavy and might not work at all depending on whether treecorr results objects can be send by MPI.

joezuntz:
Though I know that Mike is planning an MPI upgrade to TreeCorr, so the whole setup may change when that's available.

empEvil:
> 
> 
> Also, does this work under MPI? `collect_results` collects together the Measurement objects from across the processors, which would definitely be very very memory heavy and might not work at all depending on whether treecorr results objects can be send by MPI.

See that is not something I had thought about, I am not sure. Question is if there is a way to send around less than the entire TreeCorr objects, while still having a way to get the combined covariance matrix out from TreeCorr.

empEvil:
> 
> 
> Just looked at this - I think you have your covariance potentially in the wrong order. You give treecorr thing organized by ggs then ngs then nns, but that's not necessarily the order you do S.add_data_point. Is there a reason you need separate ggs, ngs, and nns?

We might actually be able to skip that step, only worry is what happens to the covariance matrix once we do the reordering step? I am not familiar enough with SACC to know if that changes anything.

joezuntz:
sacc knows how to reorder covariances along with data vectors, so that's fine.  But when you originally supply the covariance it has to match the order you did add_data_point.

empEvil:
okay then I will fix that

txpipe/twopoint.py
```diff
@@ -353,15 +410,27 @@ def get_lens_catalog(self, data, i):
...
         if 'random_bin' in data:
             random_mask = data['random_bin']==i
-            rancat  = treecorr.Catalog(
-                ra=data['random_ra'][random_mask], dec=data['random_dec'][random_mask],
-                ra_units='degree', dec_units='degree')
+            if self.config['var_methods']=='jackknife':
+                rancat  = treecorr.Catalog(
+                    ra=data['random_ra'][random_mask], dec=data['random_dec'][random_mask],
+                    ra_units='degree', dec_units='degree',
+                    npatch=self.config['npatch'])
```
rmjarvis: This won't work.  And is possibly why Joe is (sometimes?) seeing problems with the patches matching up, as mentioned on Slack.  This command makes a whole new set of patches for the random that don't correspond to the patches used for the lenses.

You want instead
```
rancat  = treecorr.Catalog(
                    ra=data['random_ra'][random_mask], dec=data['random_dec'][random_mask],
                    ra_units='degree', dec_units='degree',
                    patch_centers=cat.patch_centers)
```
This will identify the right objects in the random that correspond to each patch used by `cat`.

joezuntz: Do we need the same patches for the lens and source catalogs also, and across the different tomographic bins?

empEvil: We probably should have, if you note the initial pull request, we note that there might be a potential issue with consistency of the patches between catalogs and that needs improvement.

rmjarvis: Only if you want to make the combined covariance matrix across tomo bins.  (Which I suspect is the case!)

joezuntz: Yes, we do indeed!  @empEvil @jpratmarti - you might need to have one process generate the patches and then mpi bcast them to the others.

rmjarvis: I'd recommend restructuring to have the full catalog make the patches, and then use those patch_centers for all the tomo bins.

empEvil: My suggestion would be to add in a pipeline stage that does what Mike suggests, and I think we were planning to do that. I thought we were waiting on a newer version of treecorr though

joezuntz: I don't think you want an entirely separate stage - then it would have to read the data in all over again  Since each process has the entire data set right now anyway, having one process make the patches on the full catalog would be fine.  Not sure if that will change in future. 

Alternatively, we do have a counts map by this point made in the mapping stage, so you could use that to construct the patch centres and share them.  That could be its own stage, since loading the map is is much faster than loading the catalog.

rmjarvis: You might want to look at how I did it in this script:
https://github.com/rmjarvis/TreeCorr/pull/99/files#diff-0b1e108387239363fa6a206cf6ba9e5aR61
I load in just a portion of the full catalog for making the patches, and then write those patches to a file.  Then the other catalogs can use those patch_center files when they read in.  This may be more efficient, since it can load less data into memory when building up each patch.

empEvil:
I updated the code such that it expects a jackcenter.hdf5, to list the centers of the patches. @joezuntz How do I add the jackcenter.hdf5 file so it can be run by others?

joezuntz:
@empEvil that jackcenter file should be generated at an earlier stage of the pipeline, presumably?

empEvil:
yes indeed. I think you and @chihway had an idea of where that should go. For now I just generated the file by it self, so I and others for that matter could test the new implementation
[jackcenters.txt](https://github.com/LSSTDESC/TXPipe/files/4215710/jackcenters.txt)

empEvil:
Added such that the Diagnosticmaps stage now generates the patch_centers, by simply taking the lists of Ra and Dec from the photometry catalog, and then runs Treecorr's kmeans algorithm on them. Don't think this is the best way to do this, but now the code should be able to run it self, without extra input files

txpipe/sysmaps.py
```diff
@@ -167,6 +173,10 @@ def run(self):
         map_pix, ngals, g1, g2, var_g1, var_g2 = mapper.finalize(self.comm)
         map_pix_psf, ngals_psf, g1_psf, g2_psf, var_g1_psf, var_g2_psf = mapper_psf.finalize(self.comm)
         flag_pixs, flag_maps = flag_mapper.finalize(self.comm)
+        import treecorr
+        cat = treecorr.Catalog(ra = ra, dec = dec, ra_units='degree', dec_units = 'degree', npatch=self.config['npatch'])
+        cat.patches
```
rmjarvis: This line isn't actually necessary.  (Or shouldn't be anyway).  Calling `write_patch_centers` should be enough to trigger the patch creation.

empEvil: Okay, wasn't sure

joezuntz:
On the call just now we decided to move patch center generation to its own stage.

txpipe/sysmaps.py
```diff
@@ -126,6 +126,10 @@ def run(self):
         m_it = self.iterate_hdf('tomography_catalog','multiplicative_bias', m_cols, chunk_rows)
         m_it = (d[2] for d in m_it)
 
+        #Naive attempt at generating the Jackknife centers!
```
joezuntz, line 129: I think this is left-over code from an earlier attempt?  If so, can you remove?

txpipe/twopoint.py
```diff
@@ -234,16 +240,72 @@ def write_output(self, data, meta, results):
...
+        #nns = []
+        comb = []
         for d in results:
-            tracer1 = f'source_{d.i}' if d.corr_type in [XIP, XIM, GAMMAT] else f'lens_{d.i}'
-            tracer2 = f'source_{d.j}' if d.corr_type in [XIP, XIM] else f'lens_{d.j}'
+            #First lets try and get Covariance matrix
+            tracer1 = f'source_{d.i}' if d.corr_type in [XI, GAMMAT] else f'lens_{d.i}'
+            tracer2 = f'source_{d.j}' if d.corr_type in [XI] else f'lens_{d.j}'
+            comb.append(d.object)
+            if d.corr_type == XI:
```
joezuntz, line 252: This been repeats itself needlessly - can you move things that are the same in each branch of the `if` statement outside it?

txpipe/twopoint.py
```diff
@@ -234,16 +240,72 @@ def write_output(self, data, meta, results):
...
+                    npair = npair/2
+                    weight = weight/2
+                n = len(wtheta)
+                for i in range(n):
+                    S.add_data_point(WTHETA, (tracer1,tracer2), wtheta[i],
+                        theta=theta[i], error=wthetaerr[i], weight=weight[i])
+
             # Each of our Measurement objects contains various theta values,
             # and we loop through and add them all
-            n = len(d.value)
```
joezuntz, line 242: Delete the old code

txpipe/twopoint.py
```diff
@@ -331,12 +378,24 @@ def get_shear_catalog(self, data, i):
...
         g1,g2,mask = self.get_m(data, i)
 
-        cat = treecorr.Catalog(
-            g1 = g1,
-            g2 = g2,
-            ra = data['ra'][mask],
-            dec = data['dec'][mask],
-            ra_units='degree', dec_units='degree')
+        if self.config['var_methods']=='jackknife':
+            patchcenters = self.get_input('patch_centers')
```
joezuntz, line 382: Two of the occurrences of "patchcenters" have underscores between "patch" and "centers", yet the third does not.  Can you harmonize by changing to patch_centers? (am feeling very pedantic today)

txpipe/twopoint.py
```diff
@@ -596,12 +659,17 @@ class TXTwoPointPlots(PipelineStage):
...
+        ('shear_xi_err', PNGFile),
         ('shearDensity_xi', PNGFile),
+        ('shearDensity_xi_err', PNGFile),
         ('density_xi', PNGFile),
+        ('density_xi_err', PNGFile),
     ]
 
     config_options = {
-
+        'plot_compare_errors': True,
```
joezuntz, line 670: There's no need to make this optional, unless I'm missing something?  This will be a useful plot always.

txpipe/twopoint.py
```diff
@@ -774,10 +909,65 @@ def plot_shear_density(self, s, sources, lenses):
                     fontsize=10, verticalalignment='top', bbox=props)
 
         plt.tight_layout()
-        plt.subplots_adjust(hspace=0,wspace=0)
+        plt.subplots_adjust(hspace=self.config['hspace'],wspace=self.config['wspace'])
```
joezuntz, line 912: This also repeats a lot of the stuff above - possible to combine?

txpipe/twopoint.py
```diff
@@ -1028,17 +1288,18 @@ def load_lens_catalog(self, data):
...
         bright_cut &= mags<18.3
-        
+
         data['lens_ra']  = f['stars/ra'][:][bright_cut]
         data['lens_dec'] = f['stars/dec'][:][bright_cut]
         f.close()
 
         npoint = data['lens_ra'].size
         data['lens_bin'] = np.zeros(npoint)
+        print('len lens', len(data['lens_ra']))
```
joezuntz, line 1302: is this needed?  If so make it clearer output.

txpipe/twopoint.py
```diff
@@ -1063,17 +1324,20 @@ def write_output(self, data, meta, results):
...
         import matplotlib.pyplot as plt
         d = results[0]
+        dvalue = d.object.xi
+        derror = np.sqrt(d.object.varxi)
+        dtheta = np.exp(d.object.meanlogr)
 
         fig = self.open_output('gammat_bright_stars_plot', wrapper=True)
 
         # compute the mean and the chi^2/dof
         flat1 = 0
```
joezuntz, line 1334: Why are we subtracting off zero? Looks like this isn't from your PR, but a good place to fix it I think.

txpipe/twopoint.py
empEvil: Think This should be resolved, please check

txpipe/twopoint.py
empEvil: I just removed that line, if it is still needed feel free to add it back

txpipe/twopoint.py
empEvil: No clue, I tried to clean it up

txpipe/twopoint.py
jpratmarti: True, I removed the option.

txpipe/twopoint.py
jpratmarti: @joezuntz do you mean in general combine all the plotting code or just this line?

examples/laptop.yml
```diff
@@ -45,10 +47,8 @@ inputs:
 
 # if supported by the launcher, restart the pipeline where it left off
 # if interrupted
-resume: True
```
joezuntz, line 48: can you change this back to the default also.

txpipe/sysmaps.py
```diff
@@ -152,7 +152,7 @@ def run(self):
             shear_tmp['dec'] = phot_data['dec']
             shear_psf_tmp['ra'] = phot_data['ra']       # Does it have 'ra' ?
             shear_psf_tmp['dec'] = phot_data['dec']     # Does it have 'dec' ?
```
joezuntz, line 154: can you remove these changes so the PR as a whole is cleaner?
you can do this with git checkout to get the master version of this file.

txpipe/twopoint.py
joezuntz: The general plotting code.  No need to go crazy if it's more awkward than I thought!

jpratmarti: Yeah I was thinking about that.. I agree it'd be much better, I'll give it a try

examples/laptop.yml
jpratmarti: Done.

PR #70, opened 2020-02-26
Gaussian covariance

chihway:
This pull request is for two new stages “TXFourierGaussianCovariance” and “TXRealGaussianCovariance” to run after the associated twopoint stages and calculate analytical covariance to attach to the sacc files (and write out new files called "summary_statistics").

The code is based primarily on the earlier covariance.py file written  by @aamon and @crcmahony and the notebook from Sukhdeep in TJPCov.

The code runs, but there are several placeholders that could be improved. This should however be an ok version to run before TJPCov is fully incorporated with firecrown. The following are known issues/placeholders.

* theta bin edges should be read from sacc instead of from config (prob need to modify twopoint.py)
* IA and galaxy bias and other systematics are not included
* area calculated from self-created mask -- is that good enough?
* need to have TJPCov in path when running
* fine ell and theta grid before binning is not tested yet
* in Fourier space, B-mode covariances are not output
* coupling matrix are placeholders
* none of this is tested against other code yet
* add possibility of storing two covariances in sacc? ie. JK and analytic


joezuntz:
We resolved to make TJPCov into a package and update this to use that.

chihway:
todo before merge:

- [ ] TJP pip-installable
- [ ] check with cosmosis Gaussian covariance 
- [ ] add check on pos-def

chihway:
Ok @joezuntz I think we are ready to merge, with some caveats. There is a factor of two that @sukhdeep1989 is still checking, but otherwise it agrees to the cosmosis implementation within 10%. We can probably do better going to higher ell and finer binning, but the current settings run in a reasonable time.

Here's the notebook for the comparison:
https://github.com/LSSTDESC/txpipe-cosmodc2/blob/master/notebooks/CompareGaussianCov.ipynb
For reference, the cosmosis file to generate the cosmosis covariance is here: https://github.com/LSSTDESC/txpipe-cosmodc2/tree/master/cosmosis_config

I also added the check for pos-def,  but commented out for now since the xim will not give  a pos-def covariance  unless we use very high ells.

joezuntz:
Hi @chihway - thanks!  Two things:
- can you clean up the changes that are just spacing, etc.?  and the extra print out in selector?
- thinking about the xim thing - should we maybe manually extrapolate the C_ell out to those very high values, just so that we have something that uses all the data we have?

Cheers,
Joe

chihway:
Thanks @joezuntz 
* I cleaned up hopefully the silly stuff
* I tried to increase ell_max but still had no luck inverting xim, so for now I left the test there but outputs a warning instead of exits the code, lemme know what you think about it

@sukhdeep1989 the new code works and matches with cosmosis, I just needed to change this line `if self.do_xi and np.all(np.array(tracers)=='shear'):` to fit the actual names of the tracers.

txpipe/covariance.py
```diff
@@ -61,206 +77,419 @@ def read_cosmology(self):
...
+
+        if self.do_xi and (first_is_shear_shear or second_is_shear_shear):
+            # this adds the B-mode shape noise contribution.
+            # We assume B-mode power (C_ell) is 0
+            Bmode_F = 1
+            if xi_plus_minus1 != xi_plus_minus2:
+                # in the cross term, this contribution is subtracted.
+                # eq. 29-31 of https://arxiv.org/pdf/0708.0387.pdf
+                Bmode_F=-1 
+            cov[1324] += np.outer(cl[13]*0 + SN[13], cl[24]*0 + SN[24]) * coupling_mat[1324] * Bmode_F
```
joezuntz: Are the components here that are multiplied by zero needed to maintain the right shape? (I assume they representing are B-mode signal C_ells).  If that's right could you add a comment to that effect?

txpipe/covariance.py
```diff
@@ -61,206 +77,419 @@ def read_cosmology(self):
...
+        import threadpoolctl
+
+        #FIXME: Only input needed should be two_point_data, which is the sacc data file.
+        # Other parameters should be included within sacc and read from there.
+        ccl_tracers,tracer_Noise = self.get_tracer_info(cosmo, meta, two_point_data=two_point_data)
+        # we will loop over all these
+        tracer_combs = two_point_data.get_tracer_combinations() 
+        N2pt = len(tracer_combs)
+        
+        N2pt0 = -1
```
joezuntz: I don't understand what this bit is doing, why the counter starts at -1, or why we are multiplying things by 1.  Can you add some explicatory comments?

txpipe/covariance.py
```diff
@@ -31,23 +33,37 @@ class TXFourierGaussianCovariance(PipelineStage):
...
-        cov = self.compute_covariance(sacc_data, theory_c_ell, noise_c_ell, fsky)
-        self.save_outputs(sacc_data, cov)
-        #self.save_outputs_firecrown(C,data_vector,ells,nz,binning_info)
+        two_point_data = self.read_sacc()
+
+        # read the n(z) and f_sky from the source summary stats        
+        meta = self.read_number_statistics()
+        
+        # Binning choices. The ell binning is a linear piece with all the
+        # integer values up to 500 (is this level of accuracy needed?)
```
joezuntz: can you edit to mention that these choices cam from firecrown, and might be tweaked later?

txpipe/covariance.py
```diff
@@ -61,206 +77,419 @@ def read_cosmology(self):
...
+        ell_edges.append(X[-1]['window'].max)
+
+        return np.array(ell_edges)
+
+    #compute all the covariances and then combine them into one single giant matrix
+    def compute_covariance(self, cosmo, meta, two_point_data={}):
+        from tjpcov import bin_cov, wigner_transform
+        import threadpoolctl
+
+        #FIXME: Only input needed should be two_point_data, which is the sacc data file.
```
joezuntz: yes, you can remove this now

txpipe/covariance.py
```diff
@@ -61,206 +77,419 @@ def read_cosmology(self):
...
+        # the max value of one window is the min value of the next.
+        # So we just need the lower edges of each bin and then the
+        # final maximum value of the last bin
+        ell_edges = [x['window'].min for x in X]
+        ell_edges.append(X[-1]['window'].max)
+
+        return np.array(ell_edges)
+
+    #compute all the covariances and then combine them into one single giant matrix
+    def compute_covariance(self, cosmo, meta, two_point_data={}):
```
joezuntz: is two_point_data really an optional argument here?

txpipe/covariance.py
```diff
@@ -61,206 +77,419 @@ def read_cosmology(self):
...
-                else:
-                    assert data_type == Cdd
-                    N = np.ones(n) / n_lens[b]
-                noise[(data_type, tracer1, tracer2)] = N
+    def run(self):
+        super().run()
 
-        return noise
+    def get_angular_bins(self, two_point_data):
+        # this should be changed to read from sacc file
```
joezuntz: you're right these aren't in twopoint.py yet - will add them.

joezuntz: can leave this comment here for now

txpipe/covariance.py
chihway: yep

txpipe/covariance.py
chihway: yep

txpipe/covariance.py
chihway: ok

txpipe/covariance.py
chihway: ah no, fixed

txpipe/covariance.py
chihway: ok this bit is a bit of a hack, all it's doing is counting the number of 2pt functions and making sure xip xim are both counted... I added some comments but if you have a better way to do this please change

txpipe/covariance.py
chihway: ok

PR #72, opened 2020-02-27
Changes that make DC2 mostly work

joezuntz:
These are changes I've made while getting to some DC2 results.


joezuntz:
Actually this is way too big.  I'm going to try to split it into smaller pieces.

PR #74, opened 2020-02-27
Add fake map tool and theory support file

joezuntz:
Tool to generate fake maps for testing Fourier code

### DIFF
--- a/txpipe/sysmaps.py
+++ b/txpipe/sysmaps.py
@@ -331,6 +331,8 @@ class TXFakeMaps(TXDiagnosticMaps):
         'lens_counts':[10.0, 10.0],
         'dec_min': -70.,
         'dec_max': -10.,
+        'ra_min': 0.0,
+        'ra_max': 90.,
         'flag_exponent_max': 8,
     }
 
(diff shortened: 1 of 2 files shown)


chihway:
Looks good, only very small things:

sysmaps.py:
L306: do you actually use flask?? I think you mean healpix
L332: why only dec range?
L347: did this actually work? thought you need `from .utils.pixel_schemes import choose_pixelization`
L367: comments broken

theory.py:
L102: I think logspace is base 10 by default, so may want to specify or change to log10 in the limits
L593: comments broken

joezuntz:
TODO List:

- [ ] L306: do you actually use flask?? I think you mean healpix
- [ ] L332: why only dec range?
- [ ] L347: did this actually work? thought you need from .utils.pixel_schemes import choose_pixelization
- [ ] L367: comments broken

theory.py:
- [ ] L102: I think logspace is base 10 by default, so may want to specify or change to log10 in the limits
- [ ] L593: comments broken

joezuntz:
>  L347: did this actually work? thought you need from .utils.pixel_schemes import choose_pixelization

Yes, this works because in `utils/__init__.py` I import it from the module.

joezuntz:
Okay, I've dealt with all these - many thanks!

PR #75, opened 2020-02-27
Updated ingestion

joezuntz:
- Renames to CosmoDC2Mock and uses that catalog (instead of Proto) by default
- Updates GCR calls
- Add unit_response option to make all metacal R values either 1 (for shear R11 and R22) or 0 (everything else) 
- Adds flip_g2 option to make convention match treecorr and NaMaster
- Speed up
- Fix bug giving wrong SNR values for variants, affecting selection corrections
- Remove old code

### DIFF
--- a/txpipe/input_cats.py
+++ b/txpipe/input_cats.py
@@ -4,7 +4,7 @@ from .utils.metacal import metacal_band_variants, metacal_variants
 import numpy as np
 from .utils.timer import Timer
 
-class TXProtoDC2Mock(PipelineStage):
+class TXCosmoDC2Mock(PipelineStage):
     """
     This stage simulates metacal data and metacalibrated
     photometry measurements, starting from a cosmology catalogs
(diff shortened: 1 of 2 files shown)


txpipe/input_cats.py
```diff
@@ -351,11 +379,15 @@ def make_mock_metacal(self, data, photo):
             photo['snr_i']**2 + 
             photo['snr_z']**2
             )**0.5
-        snr_1p = (photo['snr_r_1p']**2 + photo['snr_i_1p'] + photo['snr_z_1p'])**0.5
-        snr_1m = (photo['snr_r_1m']**2 + photo['snr_i_1m'] + photo['snr_z_1m'])**0.5
-        snr_2p = (photo['snr_r_2p']**2 + photo['snr_i_2p'] + photo['snr_z_2p'])**0.5
-        snr_2m = (photo['snr_r_2m']**2 + photo['snr_i_2m'] + photo['snr_z_2m'])**0.5
-
+        snr_1p = (photo['snr_r_1p']**2 + photo['snr_i_1p']**2 + photo['snr_z_1p']**2)**0.5
```
EiffL: Good catch :-)

joezuntz: yeah, this was really messing things up!

PR #77, opened 2020-02-27
Save lensing weight map in sys maps

joezuntz:
This saves the count of the number of objects in the lensing maps as a lensing weight map for later.
Update: now saves a summed weight, though this is the same is the count in our case.

### DIFF
--- a/txpipe/mapping/basic_maps.py
+++ b/txpipe/mapping/basic_maps.py
@@ -68,6 +68,7 @@ class Mapper:
         g2 = {}
         var_g1 = {}
         var_g2 = {}
+        counts_g = {}
 
         rank = 0 if comm is None else comm.Get_rank()
         pixel = np.arange(self.pixel_scheme.npix)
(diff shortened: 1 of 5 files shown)


txpipe/mapping/basic_maps.py
```diff
@@ -123,17 +124,20 @@ def finalize(self, comm=None):
             mask[count_g1>0] = True
             mask[count_g2>0] = True
 
-            #  Don't think we want to save these at the moment
-            del count_g1
-            del count_g2
+            # We only need to keep one of these as they should be
+            # the same.
+            counts_g[b] = count_g2.reshape(self.pixel_scheme.shape).flatten()
```
EiffL: Hum.... I'm slightly confused, why are g1 and g2 collected independently? Looks like it's possible to have bins with g1 contribution but no g2 from particular galaxies....

txpipe/sysmaps.py
```diff
@@ -204,15 +204,18 @@ def run(self):
                 self.save_map(group, f"g2_{b}", map_pix, g2[b], config)
                 self.save_map(group, f"var_g1_{b}", map_pix, var_g1[b], config)
                 self.save_map(group, f"var_g2_{b}", map_pix, var_g2[b], config)
+                self.save_map(group, f"lensing_weight_{b}", map_pix, counts_g[b], config)
```
EiffL: So.... I naively would have expected "lensing_weight" to mean inverse variance weight, but if I understand correclty this is more "number of objects per pixel", right?

joezuntz:
We resolved to include a weight column and use that, even if it's 1 for metacal.

joezuntz:
@EiffL I've updated this so that it now:
- generates a lensing weight column, all ones in the ingestion stage
- calculates a summed weight map per-pixel per-tomographic bin
- saves the weight maps (as before)

Let me know if this looks alright.

txpipe/mapping/basic_maps.py
```diff
@@ -119,41 +131,42 @@ def finalize(self, comm=None):
...
-
-            v_g1 = v_g1.reshape(self.pixel_scheme.shape)
-            v_g2 = v_g2.reshape(self.pixel_scheme.shape)
-
-            g1[b] = mean_g1.flatten()
-            g2[b] = mean_g2.flatten()
+            # I really don't recall why I'm changing the shape
+            # back and forth here.  Maybe I had a reason?
+            # TODO: Remove the reshape/flatten and see what happens.
+            g1[b] = mean_g1.reshape(self.pixel_scheme.shape).flatten()
```
EiffL: This looks harmless but yeah, I don't see the reason for the reshaping, I would just try to remove the reshapes before flattening.

txpipe/mapping/basic_maps.py
joezuntz: Added

joezuntz:
I've dealt with the weird bits.  Many thanks for reviewing!

PR #78, opened 2020-02-27
Fixes to twopoint_fourier for DC2

joezuntz:
This PR fixes and updates the namaster interface.
It can only be merged after #74 because it uses the `theory_3x2pt` function from there, and after #77 because it uses the lensing weights.

Secondarily, it also adds a DC2 pipeline file.

### DIFF
--- a/txpipe/twopoint_fourier.py
+++ b/txpipe/twopoint_fourier.py
@@ -42,11 +42,14 @@ class TXTwoPointFourier(PipelineStage):
     """
     name = 'TXTwoPointFourier'
     inputs = [
-        ('photoz_stack', HDFFile),  # Photoz stack
+        ('shear_photoz_stack', HDFFile),  # Photoz stack
+        ('lens_photoz_stack', HDFFile),  # Photoz stack
         ('diagnostic_maps', DiagnosticMaps),
         ('fiducial_cosmology', YamlFile),  # For the cosmological parameters
         ('tracer_metadata', TomographyCatalog),  # For density info
         ('noise_maps', NoiseMaps),
+        ('shear_tomography_catalog', TomographyCatalog),  # For density info
+        ('lens_tomography_catalog', TomographyCatalog),  # For density info
     ]
     outputs = [
         ('twopoint_data_fourier', SACCFile)
@@ -561,21 +564,22 @@ class TXTwoPointFourier(PipelineStage):
         # but also potentially to compute the theory guess
         # for projecting out modes
         import sacc
-        f = self.open_input('photoz_stack')
+        f_shear = self.open_input('shear_photoz_stack')
+        f_lens = self.open_input('lens_photoz_stack')
 
         tracers = {}
 
         for i in range(nbin_source):
             name = f"source_{i}"
-            z = f['n_of_z/source/z'][:]
-            Nz = f[f'n_of_z/source/bin_{i}'][:]
+            z = f_shear['n_of_z/source/z'][:]
+            Nz = f_shear[f'n_of_z/source/bin_{i}'][:]
             T = sacc.BaseTracer.make("NZ", name, z, Nz)
             tracers[name] = T
 
         for i in range(nbin_lens):
             name = f"lens_{i}"
-            z = f['n_of_z/lens/z'][:]
-            Nz = f[f'n_of_z/lens/bin_{i}'][:]
+            z = f_lens['n_of_z/lens/z'][:]
+            Nz = f_lens[f'n_of_z/lens/bin_{i}'][:]
             T = sacc.BaseTracer.make("NZ", name, z, Nz)
             tracers[name] = T
 
--- a/txpipe/twopoint_fourier.py
+++ b/txpipe/twopoint_fourier.py
@@ -398,7 +398,7 @@ class TXTwoPointFourier(PipelineStage):
         # The healpix pixel windows.  C_ell estimates on healpix
         # maps (like all pixelized maps) are modulated by a pixel
         # window function, which we calculate here so we can remove
-        # it below.  These are trivially fast to compute, so no point
+        # it below.  These are trivially fast to get, so no point
         # caching.
         if (pixel_scheme.name == 'healpix') and self.config['pixel_window']:
             # Get the raw pixel window functions
@@ -444,7 +444,7 @@ class TXTwoPointFourier(PipelineStage):
         workspace.compute_coupling_matrix(field_i, field_j, ell_bins)
 
         # Get the coupled noise C_ell values to give to the master algorithm
-        cl_noise = self.compute_noise(i,j,k,ell_bins,workspace,tomo_info)
+        cl_noise = self.compute_noise(i, j, k, ell_bins, workspace, tomo_info)
 
         # Run the master algorithm
         if pixel_scheme.name == 'healpix':
@@ -488,7 +488,6 @@ class TXTwoPointFourier(PipelineStage):
 
         # Couple to take coupled C_ell value
         if ell_bins.is_flat():
-            ell = ell_bins.get_effective_ells()
             cl_noise = w.couple_cell(N2)
         else:
             cl_noise = w.couple_cell(N2)
--- a/txpipe/twopoint_fourier.py
+++ b/txpipe/twopoint_fourier.py
@@ -2,11 +2,13 @@ from .base_stage import PipelineStage
 from .data_types import MetacalCatalog, TomographyCatalog, RandomsCatalog, YamlFile, SACCFile, DiagnosticMaps, HDFFile, PhotozPDFFile
 import numpy as np
 import collections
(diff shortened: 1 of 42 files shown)


txpipe/twopoint_fourier.py
```diff
@@ -157,111 +156,128 @@ def load_maps(self):
...
-            ng[cut] = 0
+            clustering_weight[ng==healpy.UNSEEN] = 0
 
+        d_maps = []
+        for ng in ngal_maps:
+            # Convert the number count maps to overdensity maps.
+            # First compute the overall mean object count per bin.
+            # Maybe we should do this in the mapping code itself?
+            # mean clustering galaxies per pixel in this map
+            mu = ng[clustering_weight>0].mean()
```
damonge, line 234: is this a weighted mean?

txpipe/twopoint_fourier.py
```diff
@@ -157,111 +156,128 @@ def load_maps(self):
...
+        # And then the lensing maps
+        for i, (g1, g2, lw) in enumerate(zip(g1_maps, g2_maps, lensing_weights)):
+            if self.rank == 0:
+                print(f"Generating lensing field {i}")
+            # Same again.  NaMaster knows whether these are spin 0 or spin 2
+            # from the length of the maps list we pass it (1 or 2).
+            if pixel_scheme.name == 'gnomonic':
+                field = nmt.NmtFieldFlat(lx, ly, lw, [g1,g2], templates=syst_wl) 
+            elif pixel_scheme.name == 'healpix':
+                field = nmt.NmtField(lw, [g1, g2], templates=syst_wl)
```
damonge, line 274: So, just to clarify:
- `lw` contains the sum of weights in each pixel
- `g1/2` contain the weighted mean of gamma_1/2 in each pixel (or zero if the sum of weights in that pixel is zero)
correct?

txpipe/twopoint_fourier.py
damonge: Actually, what is `mu` exactly? Is it a map?
In principle each pixel should contain:
```
delta[i] = number[i] / (mean_number * mask[i]) - 1
```
where `number` is the number of objects in that pixel, `mask` is the effective area fraction of that pixel (is this what you're calling `clustering_weight`?) and `mean_number = np.sum(number * mask) / np.sum(mask)`.

txpipe/twopoint_fourier.py
```diff
@@ -348,137 +358,177 @@ def select_calculations(self, nbins_source, nbins_lens):
...
+            pixwin_ell = np.arange(pixwin_t.size)
+            pixwin_t = np.interp(ls, pixwin_ell, pixwin_t)
+            pixwin_p = np.interp(ls, pixwin_ell, pixwin_p)
+        else:
+            # TODO figure out what these are supposed to be
+            pixwin_t = 1.0
+            pixwin_p = 1.0
+
+        # We need the theory spectrum for this pair
+        #TODO: when we have templates to deproject, use this.
```
damonge, line 416: Just checking: you've not deprojected anything in the plots you sent me on slack, right?

txpipe/twopoint_fourier.py
```diff
@@ -348,137 +358,177 @@ def select_calculations(self, nbins_source, nbins_lens):
...
+
+        if pixel_scheme.name == 'healpix':
+            workspace = nmt.NmtWorkspace()
+        elif pixel_scheme.name == 'gnomonic':
+            workspace = nmt.NmtWorkspaceFlat()
+        else:
+            raise ValueError(f"No NaMaster workspace for pixel scheme {pixel_scheme.name}")
+
+        # Compute mode-coupling matrix
+        workspace.compute_coupling_matrix(field_i, field_j, ell_bins)
```
damonge, line 444: Just a reminder that it's possible to save these to file and read them if you want to avoid having to recalculate them every time.

damonge: Also, if masks are the same (as may usually be the case for clustering), then you don't need to recompute things.

txpipe/twopoint_fourier.py
```diff
@@ -348,137 +358,177 @@ def select_calculations(self, nbins_source, nbins_lens):
...
             return None
-        # print("Setting noise to zero")
-        # return None
+
         # We loaded in sigma_e and the densities
-        # earlier on and put them in the noise dictionary
-        noise_level = noise[(i,k)]
+        # earlier on, and put them in the tomo_info dictionary
+        if k==SHEAR_SHEAR:
+            noise_level = tomo_info['sigma_e'][i]**2 / tomo_info['n_eff_steradian'][i]
```
damonge, line 471: OK, so this is definitely problematic. The best way to do this would be to do the randomly-rotated ellipticity trick I mentioned. Have you tried that? If you want to apply an analytical formula (that I think will fail as soon as shape noise is spatially varying), I think Andrina had one, so I can ask her.

txpipe/twopoint_fourier.py
```diff
@@ -348,137 +358,177 @@ def select_calculations(self, nbins_source, nbins_lens):
...
+        else:
+            noise_level = 1.0 / tomo_info['n_lens_steradian'][i]
```
damonge, line 473: Same here. You should make this `np.mean(mask) / tomo_info['n_lens_steradian']` (assuming `tomo_info['n_lens_steradian']` was computed as `np.sum(number_map * mask) / np.sum(mask)` as I mentioned above).

txpipe/twopoint_fourier.py
```diff
@@ -348,137 +358,177 @@ def select_calculations(self, nbins_source, nbins_lens):
...
-        # # Need the same noise for EE, BB, EB, BE
-        # # or PE, PB
         if k==SHEAR_SHEAR:
             N2 = [N1, N1, N1, N1]
-        elif k==SHEAR_POS:
-            N2 = [N1, N1]
         else:
             N2 = [N1]
 
+        # Couple to take coupled C_ell value
```
damonge, line 489: If you do the above, then you shouldn't do this

damonge: I.e. just return `N2` as `cl_noise`.

txpipe/twopoint_fourier.py
```diff
@@ -157,111 +156,128 @@ def load_maps(self):
...
         # Mask any pixels which have the healpix bad value
-        for m in g1_maps + g2_maps + ngal_maps:
-            mask[m==healpy.UNSEEN] = 0
+        for (g1, g2, lw) in zip(g1_maps, g2_maps, lensing_weights):
+            lw[g1 == healpy.UNSEEN] = 0
+            lw[g2 == healpy.UNSEEN] = 0
+
+        # When running on the CosmoDC2 mocks I've found I need to flip
+        # both g1 and g2 in order to get both positive galaxy-galaxy lensing
+        # and shear-shear spectra.
```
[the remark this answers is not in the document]
joezuntz: Yeah - there's no consistency.  Metacal and CosmoDC2 use different conventions.  I think these options will always be needed!

txpipe/twopoint_fourier.py
```diff
@@ -157,111 +156,128 @@ def load_maps(self):
...
+        # I'm thinking of deleting support for this since it adds a whole
+        # set of options to maintain.
+        if pixel_scheme.name == 'gnomonic':
+            lx = np.radians(pixel_scheme.size_x)
+            ly = np.radians(pixel_scheme.size_y)
+
+        # For the lensing mask we optionall apodize the mask.
+        # TODO: include the masked object fraction in here.
+        # TODO: ask about including the depth map in here
+        if apod_size > 0:
```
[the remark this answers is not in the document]
joezuntz: Not for lensing.  Sometimes for clustering.  I will remove if you think it's never likely to be useful?

txpipe/twopoint_fourier.py
```diff
@@ -157,111 +156,128 @@ def load_maps(self):
...
 
         map_file.close()
 
-        # Cut out any pixels below the threshold,
-        # zeroing out any pixels there
-        cut = mask < mask_threshold
-        mask[cut] = 0
+        # THe gnomonic (tangent plane) mapping versions need the map width
+        # I'm thinking of deleting support for this since it adds a whole
+        # set of options to maintain.
```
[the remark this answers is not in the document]
joezuntz: Will leave it in for now.

txpipe/twopoint_fourier.py
joezuntz: Nope.

txpipe/twopoint_fourier.py
joezuntz: Yep - will definitely implement that once we have multiple lens bins.

txpipe/twopoint_fourier.py
joezuntz: I'd hoped this would work for a uniform shape noise that we have here, but if not then I'll implement the rotations.

Can I rotate an NmtField in-place, or do I need to make a new one for each rotation?

PR #79, opened 2020-03-12
Bright Object mask

joezuntz:


joezuntz:
I'm just generating a new small version of the example data set with extendedness included to test this on.  If that all works I'll merge.

joezuntz:
Thanks!

Issue #65, opened 2020-01-24
Add multiple (external) lens bins


chihway:
We want to have the capability of using e.g. redmagic for the lenses. That is, 1) there are multiple bins and 2) they come with given redshift 3) txpipe config should include parameters that specify the lens redshift binning

fjaviersanchez:
@joezuntz would it be sensible to split `TXSelector` in two separate tasks `TXSourceSelection` and `TXLensSelection`?

aimalz:
I haven't thought about how it propagates into TXPipe design, but I think the plan for [the project to make a pipeline element that optimizes bin assignments](https://confluence.slac.stanford.edu/display/LSSTDESC/TXSelector%3A+optimizing+the+tomographic+binning+strategy) could do both simultaneously.

joezuntz:
@fjaviersanchez sorry not to reply to this ages ago - yes, splitting into two is useful.  @aimalz , I think the step you're describing can be developed in parallel to this and may replace them later.  

At the very least the selectors need to choose which objects can be considered for different lens tomo and source bins.

chihway:
separate into two scripts: 
lens_selector.py, source_selector.py, each with its own class

source (@chihway):
- not much change, just the option of non-mcal = input pz
- test random forest performance

lens (@fjaviersanchez ):
- similar option of input pz
- keep cmass/lowz? as an option, add more general stuff (mag-lim etc)
- extend to include more lens bins

PR #82, opened 2020-04-13
Stages/selector split

fjaviersanchez:
This PR splits the selector stage in lens and source selector, trying to add flexibility to TXPipe


EiffL:
Hi guys, just looking through the TXPipe repo to get the lay of the land. It this PR meant to address #65 ?

chihway:
Oh yeah it's linked to that. But I think it will only partially address stuff in that issue. For now we just want to split the two stages and make sure we can do >1 lens bins :P, but I don't think after this is merged we would definitely get to use external catalogs automatically.

EiffL:
ok thanks @chihway :-) I've linked it to that issue, keeping in mind that wont fix everything

chihway:
@joezuntz @fjaviersanchez ok this now runs through all normal stages (I haven't tested stages that are not in the laptop.yml).

Most of the changes are just plumbing due to the splitting of the lens and source selector/files. Comments welcome as always.

As usual, probably can use some more checking... 

Oh and I changed the file name sysmaps.py, which is why it's complaining above.

txpipe/random_cats.py
```diff
@@ -8,8 +8,8 @@ class TXRandomCat(PipelineStage):
     name='TXRandomCat'
     inputs = [
         ('diagnostic_maps', DiagnosticMaps),
-        ('tomography_catalog', TomographyCatalog),
-        ('photoz_stack', HDFFile),
+        ('shear_tomography_catalog', TomographyCatalog),  #WHY?
```
joezuntz, line 11: We give the random objects ellipticities too, hence using that catalog.  But it can just use tracer_metadata instead, since it just needs sigma_e.  Could you change this?

examples/laptop.yml
```diff
@@ -42,11 +44,14 @@ config: examples/config.yml
...
+    photometry_catalog:  /global/cscratch1/sd/jsanch87/TXPipe_data/photometry_catalog_small.h5
+    #shear_catalog: data/example/inputs/shear_catalog.hdf5
+    #photometry_catalog: data/example/inputs/photometry_catalog.hdf5
+    photoz_trained_model: /global/cscratch1/sd/jsanch87/TXPipe_data/cosmoDC2_trees_i25.3.npy
+    calibration_table: /global/cscratch1/sd/jsanch87/TXPipe_data/sample_cosmodc2_w10year_errors.dat
+    exposures: /global/cscratch1/sd/jsanch87/TXPipe_data/exposures.hdf5
+    #star_catalog: /global/cscratch1/sd/jsacnh87/TXPipe_data/star_catalog.hdf5
     star_catalog: data/example/inputs/star_catalog.hdf5
     # This file comes with the code
     fiducial_cosmology: data/fiducial_cosmology.yml
```
joezuntz, line 57: change everything in this file back to what it was except for the selector bits

txpipe/lens_selector.py
```diff
@@ -0,0 +1,271 @@
...
+        original_warning_settings = np.seterr(all='ignore')  
+
+        # The output file we will put the tomographic
+        # information into
+        output_file = self.setup_output()
+
+        # various config options
+        chunk_rows = self.config['chunk_rows']
+
+        phot_cols = ['i_mag','r_mag','g_mag', 'redshift_true']
```
joezuntz, line 68: Can you print out a note here that we are cheating and using the true redshift?

txpipe/photoz_stack.py
```diff
@@ -60,18 +62,32 @@ def run(self):
             self.config['chunk_rows']  # number of rows to read at once
         )
 
-        tomography_iterator = self.iterate_hdf(
-            'tomography_catalog', # tag of input file to iterate through
+        photoz_iterator2 = self.iterate_hdf(
```
joezuntz, line 65: it looks like you're looping through all the data twice here, which is not needed.  Loading the entire data set is quite slow, so we try to minimize it.  You can combine these in the same loop.

txpipe/photoz_stack.py
```diff
@@ -306,17 +348,24 @@ def run(self):
             self.config['chunk_rows']  # number of rows to read at once
         )
 
-        tomography_iterator = self.iterate_hdf(
```
joezuntz, line 309: same here - you don't need two loops.

txpipe/source_selector.py
```diff
@@ -0,0 +1,431 @@
...
+        return sel
+
+
+def flatten_list(lst):
+    return [item for sublist in lst for item in sublist]
+
+
+if __name__ == '__main__':
+    PipelineStage.main()
+
```
joezuntz, line 431: Can you let me know on the PR page if you made any changes to this, other than splitting it off into its own file?

txpipe/random_cats.py
chihway: done

examples/laptop.yml
chihway: done

txpipe/lens_selector.py
chihway: done

txpipe/photoz_stack.py
chihway: ahhh, ok I think I understand that now, changed

txpipe/photoz_stack.py
chihway: changed

txpipe/source_selector.py
chihway: don't think so, yeah I'll add to the PR page

chihway:
Summary of the content of this PR:

* This PR splits the source and lens selection into two stages. 
* The source selection itself is essentially unchanged from before, while the lens selection now uses the *true* redshift and splits into different tomographic bins on top of some lens selection (currently CMASS+LOWZ selection). 
* Both source and lens selection still has lots of room for improvement.
* The resulting change in the file structure is that now there are two tomography files and two photoz_stack files. 
* For the code there are now two Selector classes and two NumberDensityStats classes.

joezuntz:
I've made some requests, and also could you merge in the master branch and get the CI working?

examples/laptop.yml
```diff
@@ -42,8 +44,10 @@ config: examples/config.yml
 
 inputs:
     # See README for paths to download these files
-    shear_catalog: data/example/inputs/shear_catalog.hdf5
-    photometry_catalog: data/example/inputs/photometry_catalog.hdf5
+    shear_catalog:  /global/cscratch1/sd/jsanch87/TXPipe_data/shear_catalog_small.h5
```
joezuntz, line 47: looks like these are still changed.

chihway: right, so the photometry catalog now needs a true redshift, and the shear catalog needs the metacal version of the redshfits... I think the right thing to do is to change the example input files. how do I change that?

chihway:
Reply above -- I think the example input should change, lemme know what to do there.

Also merged master into this branch. Most of the changes seem to be the travis stuff, what do you mean to have the Cl working?

chihway:
err ok things are failing after merge, lemme fix it first

Issue #84, opened 2020-04-20
Failure of TXSelector at NERSC from git diff


EiffL:
For some reason TXSelector failed for me multiple times when running inside a shifter interactive session at NERSC. I got the following error:
```
Executing stage: TXSelector
Traceback (most recent call last):
  File "/usr/lib/python3.6/subprocess.py", line 425, in run
    stdout, stderr = process.communicate(input, timeout=timeout)
  File "/usr/lib/python3.6/subprocess.py", line 863, in communicate
    stdout, stderr = self._communicate(input, endtime, timeout)
  File "/usr/lib/python3.6/subprocess.py", line 1535, in _communicate
    self._check_timeout(endtime, orig_timeout)
  File "/usr/lib/python3.6/subprocess.py", line 891, in _check_timeout
    raise TimeoutExpired(self.args, orig_timeout)
subprocess.TimeoutExpired: Command '['git', 'diff', 'HEAD']' timed out after 5 seconds
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/global/u2/f/flanusse/repo/TXPipe/txpipe/__main__.py", line 5, in <module>
    PipelineStage.main()
  File "/usr/local/lib/python3.6/dist-packages/ceci/stage.py", line 604, in main
    stage.execute(args)
  File "/usr/local/lib/python3.6/dist-packages/ceci/stage.py", line 679, in execute
    stage.run()
  File "/global/u2/f/flanusse/repo/TXPipe/txpipe/selector.py", line 82, in run
    output_file = self.setup_output()
  File "/global/u2/f/flanusse/repo/TXPipe/txpipe/selector.py", line 328, in setup_output
    outfile = self.open_output('tomography_catalog', parallel=True)
  File "/global/u2/f/flanusse/repo/TXPipe/txpipe/base_stage.py", line 84, in open_output
    extra_provenance = self.gather_provenance()
  File "/global/u2/f/flanusse/repo/TXPipe/txpipe/base_stage.py", line 27, in gather_provenance
    provenance["gitdiff"] = git_diff()
  File "/global/u2/f/flanusse/repo/TXPipe/txpipe/utils/provenance.py", line 53, in git_diff
[...]

PR #85, opened 2020-04-21
Provenance fixes

joezuntz:
Avoid crashes in provenance tracking when on very slow file systems or when encountering other git problems, or missing provenance in the input data files.

Addresses #84


joezuntz:
@ellongley  - this should hopefully fix your issue - could you have a look and see if it all looks okay?

ellongley:
Yes this ran great! 👍

PR #93, opened 2020-06-04
Version that works with 2.2i and CosmoDC2

joezuntz:
Includes:
- Reshuffle of yml files etc
- tool to combine iterators
- diagnostic plot improvements
- PSF-related plots use choice of PSF columns
- Ingest non-mcal PSFs
- rename of columns to match correct names
- PDF lens selector
- nicer plotting
- fixes to use split tomo cats


txpipe/diagnostic_maps.py
```diff
@@ -60,6 +62,7 @@ class TXDiagnosticMaps(PipelineStage):
         'true_shear' : False,
         'flag_exponent_max': 8,
         'dilate': True,
+        'psf_prefix': 'psf_',
```
chihway, line 65: are you imagining there to be flexibility here? if we are compiling the input shear catalog can't we always call it a fix name?

txpipe/diagnostic_maps.py
```diff
@@ -390,6 +394,28 @@ def copy_attrs(tomo, name, out_name):
...
+            'lens_density': lens_density.tolist(),
+            'source_density': source_density.tolist(),
+            'sigma_e': shear_tomo_file['tomography/sigma_e'][:].tolist(),
+            'n_eff': n_eff.tolist(),
+            'R_gamma_mean': shear_tomo_file['metacal_response/R_gamma_mean'][:].tolist(),
+            'R_S': shear_tomo_file['metacal_response/R_S'][:].tolist(),
+            'R_total': shear_tomo_file['metacal_response/R_total'][:].tolist(),
+            'lens_counts': lens_counts.tolist(),
+            'source_counts': source_counts.tolist(),
+            'area': float(area),
```
chihway, line 409: it's probably obvious, but do we want to store the units?

txpipe/diagnostic_maps.py
```diff
@@ -390,6 +394,28 @@ def copy_attrs(tomo, name, out_name):
 
         meta_file.close()
 
+        # human readable version
+        yaml_out_name = self.get_output('tracer_metadata_yml')
+        metadata = {
+            'lens_density': lens_density.tolist(),
```
chihway, line 400: do these come with units?

txpipe/diagnostics.py
```diff
@@ -25,12 +27,14 @@ class TXDiagnosticPlots(PipelineStage):
...
         ('mag_hist', PNGFile),
+        ('response_hist', PNGFile),
 
     ]
 
     config_options = {
         'chunk_rows': 100000,
-        'delta_gamma': 0.02
+        'delta_gamma': 0.02,
+        'psf_prefix': 'mcal_psf_',
```
chihway, line 37: ok now I see that there are different PSF column names... this will probably be fixed once we merge the shear calibrator branch from Emily

txpipe/diagnostics.py
```diff
@@ -455,6 +471,84 @@ def plot_snr_histogram(self):
...
+        # with extremely high values
+        edges = np.linspace(-3, 3, size+1)
+        mid = 0.5*(edges[1:] + edges[:-1])
+        width = edges[1] - edges[0]
+        # count of objects
+        counts = np.zeros((2,2,size))
+        # make a separate histogram of the shear-sample-selected
+        # objects
+        counts_s = np.zeros((2,2,size))
+        while True:
```
chihway, line 489: This might just be me being dumb, but what does this do?

txpipe/diagnostics.py
```diff
@@ -455,6 +471,84 @@ def plot_snr_histogram(self):
...
+            data = yield
+
+            if data is None:
+                break
```
chihway, line 493: This might just be me being dumb, but what does this do?

txpipe/photoz_stack.py
```diff
@@ -317,7 +317,7 @@ def run(self):
         # Data we need - the photometry catalog for DC2 has the true redshift
         # value in, and the tomography catalog has the binning.
         photo_iterator = self.iterate_hdf(
-            'photometry_catalog', # tag of input file to iterate through
+            'shear_photometry_catalog', # tag of input file to iterate through
```
chihway, line 320: I missed this, so we are separating the photometry catalogs? also we should probably unify what we want to use sources or shear :)

chihway:
@joezuntz  sorry to be so very slow… I went through the PR, have some very minor comments inline, but otherwise looks fine to me. You mentioned there’s a problem with fit_straightline? I couldn’t see anything wrong, what exactly was the issue?

txpipe/diagnostic_maps.py
joezuntz: Yep, exactly, as you mention below.  It's mainly because mcal_psf is a rounded version which isn't that useful for diagnostics.

txpipe/diagnostic_maps.py
joezuntz: I'm refactoring the mapping code in another branch - will add that there.

txpipe/diagnostic_maps.py
joezuntz: added in branch redmagic_lens

txpipe/diagnostics.py
joezuntz: These are co-routines - functions that you can pause and come back to later (in this case when we load the next chunk of data.  There's an explanation in the run method, and I'll talk about them at the remote meeting DE school in July!

txpipe/photoz_stack.py
joezuntz: mistake - fixed!

PR #94, opened 2020-06-08
Fixes to work in Fourier space with 2.1i data

fjaviersanchez:
This PR contains several small fixes to be able to run the `TXTwoPointFourier` stage in 2.1i data

### DIFF
--- a/txpipe/base_stage.py
+++ b/txpipe/base_stage.py
@@ -20,8 +20,9 @@ class PipelineStage(PipelineStageBase):
             provenance[f"config/{key}"] = str(value)
 
         for name, tag_cls in self.inputs:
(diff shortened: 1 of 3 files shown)


fjaviersanchez:
Most of the changes are in `TXNoiseMaps`. I think that these changes may be backwards incompatible though.

fjaviersanchez:
I just saw that a lot of these fixes are already in #93

joezuntz:
Hey - thanks for these!  I've manually put the first two fixes in place in the master branch.  The latter one won't work in the case where the lens branch is external, but since I'm in the process of fixing that anyway for the redmagic I'll deal with it separately.  So I'm going to close this, but only because I'm copying it to fix elsewhere!

Issue #71, opened 2020-02-27
TODO list for plots and null tests


joezuntz:
Early Tests & Plots
-------------------

These tests can be done at the start of the pipeline, just by looking at input data.

They all focus on the PSF.  There should be three star samples:
- PSF stars
- Reserved stars
- All certain stars

Tests we need are:

- [x] Rowe statistics
- [ ] PSF residual whisker plot (reserved stars) across FoV
- [x] Histogram of PSF size residuals
- [x] PSF size residual as function of magnitude

Intermediate Tests & Plots
--------------------------

These tests need some kind of calculated TXPipe intermediate or final quantities, often just the list of galaxies selected, sometimes tomographic bins, and sometimes other quantities.  Many of them require the reponse factors to be correctly included.

- [x] Alpha/beta fit of shear to PSF
- [ ] Mean shear plot across FoV
- [ ] Gamma_t/x around chip centres
- [x] Gamma_t/x around field centres
- [x] Gamma_t/x around stars
- [x] Gamma_t/x around random points
- [x] Correlations between lensing map and systematics maps
- [x] Dependence of source density and ellipticity on observing conditions
- [x] Dependence of source density and ellipticity on n_visit
- [x] Star-Galaxy correlation function
- [x] B-modes
- [x] B-mode mass aperture statistic
- [x] gamma_x around lenses
- [x] Difference in corr functions when split by S/N
- [x] Difference in corr functions when split by systematic quantities
- [x] Difference in corr functions when split by chip location
- [x] Difference in corr functions when split by chip number

joezuntz:
It's ocurred to me that the best way to do correlation functions with size and SNR splits is to just re-run the entire pipeline with different cuts, since that will do everything including computing the new n(z) values.  So I'll call that complete.

joezuntz:
I think we've concluded that FOV/chip plots are not that useful for LSST since the number of exposures is so much higher so everything averages down. So the remaining plots will not be that useful, and I'll close this ancient issue now!

PR #101, opened 2020-06-16
adding TXGammaTRandoms stage

ih64:
I mostly copy pasted TXGammaDimStars and then changed it around so it loads the random catalog instead of the star catalog. Not sure if I handled the sacc outputs correctly, but happy to go back and make changes


ih64:
forgot to mention this is aimed at issue #71 , one of the items in the issue is to do gamma_t/x around randoms

txpipe/twopoint.py
```diff
@@ -1429,6 +1429,164 @@ def write_output_sacc(self, data, meta, results):
...
         # Also make a plot of the data
 
+class TXGammaTRandoms(TXTwoPoint):
+    """
+    This subclass of the standard TXTwoPoint uses the centers
+    of stars as "lenses", as a systematics test.
+    """
+    name = "TXGammaTRandoms"
+    inputs = [
+        ('shear_catalog', MetacalCatalog),
```
joezuntz, line 1439: I think some of these aren't needed for this?

txpipe/twopoint.py
```diff
@@ -1429,6 +1429,164 @@ def write_output_sacc(self, data, meta, results):
...
+        'calcs':[0,1,2],
+        'min_sep':2.5,
+        'max_sep':100,
+        'nbins':20,
+        'bin_slop':0.1,
+        'sep_units':'arcmin',
+        'flip_g2':True,
+        'cores_per_task':20,
+        'verbose':1,
+        'reduce_randoms_size':1.0,
```
joezuntz, line 1463: Are we still subtracting randoms?  Not sure if that makes sense in this context!  This will be in the superclass.

joezuntz:
Just asked a couple of questions - thanks1

PR #103, opened 2020-06-18
Map refactor

joezuntz:
Rewrite the mapping code to split into source, lens, aux, and mask.
This will allow using external catalogs as map inputs.


examples/config/2.2i_config.yml
```diff
@@ -66,21 +66,20 @@ TXTwoPoint:
...
 
-TXDiagnosticMaps:
-    chunk_rows: 100000  # Number of rows to process at once
-    sparse: True  # Generate sparse maps - faster if using small areas
-    snr_threshold: 10.0
-    snr_delta: 1.0
+TXMainMaps:
     pixelization: healpix
     nside: 1024
-    # pixelization: gnomonic
```
chihway, line 76: just to clarify -- so we are fully healpix now, is that right? no for flat-sky possibilities?

examples/config/cosmodc2_config.yml
```diff
@@ -16,6 +16,22 @@ TXCosmoDC2Mock:
...
 
 
+TXMainMaps:
+    chunk_rows:  100000
+    pixelization: healpix
+    nside: 512
+    sparse: True
+
+TXAuxiliaryMaps:
+    chunk_rows:  100000
```
chihway, line 26: I wonder if we want to set this a variable within the config file since it's repeated so many times!

examples/config/cosmodc2_config.yml
```diff
@@ -16,6 +16,22 @@ TXCosmoDC2Mock:
...
+    sparse:  True
```
chihway, line 27: also, only the main map has the parameter "pixelization", why?

examples/laptop_complete_pipeline.yml
```diff
@@ -6,7 +6,12 @@ stages:
     - name: FlexZPipe
       #- name: PZPDFMLZ
     - name: TXPhotozStack
-    - name: TXDiagnosticMaps
+    - name: TXMainMaps
```
chihway, line 9: This is general for all the stages -- now that we are accumulating more and more stages and it's less clear what each stages does from the name. Maybe put a short comment after each of them here?

txpipe/auxiliary_maps.py
```diff
@@ -0,0 +1,200 @@
...
+        "chunk_rows": 100_000,
+        "sparse": True,
+        "flag_exponent_max": 8,  # flag bits go up to 2**8 by default
+        "psf_prefix": "psf_",  # prefix name for columns
+        "bright_obj_threshold": 22.0,  # The magnitude threshold for a object to be counted as bright
+        "depth_band": "i",  # Make depth maps for this band
+        "snr_threshold": 10.0,  # The S/N value to generate maps for (e.g. 5 for 5-sigma depth)
+        "snr_delta": 1.0,  # The range threshold +/- delta is used for finding objects at the boundary
+    }
+    # instead of reading from config we match the basic maps
```
chihway, line 39: ah ok I see now that you're matching the basic -- so there's no scenarios where we just want to run this without the basic maps?

examples/config/cosmodc2_config.yml
joezuntz: Yeah, that's a good idea. You can use the section "global" for configuration that (if needed) goes to every stage, and I've pushed changes to do that.

txpipe/auxiliary_maps.py
```diff
@@ -0,0 +1,200 @@
...
+        psf_prefix = self.config["psf_prefix"]
+        shear_catalog_type = read_shear_catalog_type(self)
+
+        # Flag column name depends on catalog type
+        if shear_catalog_type == "metacal":
+            shear_cols = [f"{psf_prefix}g1", f"{psf_prefix}g2", "mcal_flags", "weight"]
+        else:
+            shear_cols = [f"{psf_prefix}g1", f"{psf_prefix}g2", "flags", "weight"]
+
+        # See maps.py for an explanation of this
```
chihway, line 95: eh, where is maps.py? I can't seem to find it... and yeah I'm confused what this bit is doing.

txpipe/data_types/types.py
```diff
@@ -27,6 +31,23 @@ def read_catalog_info(self):
         shear_catalog_type = info.get('catalog_type')
         return shear_catalog_type
 
+    @property
```
chihway, line 34: just curious, what does this do?

txpipe/data_types/types.py
```diff
@@ -54,19 +75,8 @@ class RandomsCatalog(HDFFile):
```
chihway, line 75: does the random files have to have e1/e2? seems strange to require e.g. randoms for clustering to have e1/e2

txpipe/map_plots.py
```diff
@@ -0,0 +1,134 @@
...
+
+class TXMapPlots(PipelineStage):
+    """
+    Make plots of all the available maps.
+    """
+
+    name = "TXMapPlots"
+
+    inputs = [
+        ("source_maps", MapsFile),
```
chihway, line 13: probably don't need to do this now -- but at some point we should have the flexibility to run just subsets of the 2pt  functions and therefore not need to have all the maps

txpipe/auxiliary_maps.py
chihway: ooooh, found it!

txpipe/maps.py
```diff
@@ -0,0 +1,503 @@
...
+
+    inputs = [
+        ("lens_catalog", HDFFile),
+        ("lens_tomography_catalog", TomographyCatalog),
+    ]
+
+    config_options = {**map_config_options}
+
+    def data_iterator(self):
+        print("TODO: no lens weights here")
```
chihway, line 325: why no weights here? or do you mean you're gonna add it later?

txpipe/metadata.py
```diff
@@ -0,0 +1,124 @@
...
+        yaml.dump(metadata, f)
+        f.close()
+
+        shear_tomo_file.close()
+        lens_tomo_file.close()
+
+    def read_area(self):
+        with self.open_input("mask", wrapper=True) as f:
+            m = f.read_map("mask")
+        print(m.max())
```
chihway, line 120: remove or add something to say what this number is

txpipe/metadata.py
```diff
@@ -0,0 +1,124 @@
...
+        f_sky = (m > 0).sum() / m.size
+        area_sq_deg = f_sky * 41252.96125
+        print(area_sq_deg)
```
chihway, line 123: remove or add something to say what this number is

txpipe/source_selector.py
```diff
@@ -74,6 +73,9 @@ def run(self):
         # Suppress some warnings from numpy that are not relevant
         original_warning_settings = np.seterr(all='ignore')  
 
+        # Are we using a metacal or lensfit catalog?
```
chihway, line 76: do we want the code to decide itself what kind of shear catalog it is instead of setting it as a config option? could you explain what is the rational?

chihway:
@joezuntz ok I left some comments. I went through it pretty quick so if there's something specific you want me to look in more detail lemme know.

examples/config/2.2i_config.yml
joezuntz: Thanks - this is a good point - I think it might be handy to at least make diagnostic and other maps in gnomonic, so I've pushed some changes to make that work properly.

l have (for now) left this out of TXTwoPointFourier stages - an informative error message is generated if you try.

txpipe/data_types/types.py
joezuntz: @property is a general python decorator that turns a class method that you have to call into an attribute that you can just get with a dot, so in this case you can just do  `t = shear_cat.catalog_type` not e.g. `t = shear_cat.get_catalog_type()`.   Here's a nice description: https://www.programiz.com/python-programming/property

txpipe/data_types/types.py
joezuntz: No, it doesn't, good point.  If we do need shear randoms later this won't work because we need to use a different tomography catalog.

I've pushed a change to fix this.

txpipe/map_plots.py
joezuntz: Agreed!  I've added a note in twopoint.

txpipe/maps.py
joezuntz: Yes, meant it was to add later.  I've updated for clarity

txpipe/metadata.py
joezuntz: Done.

txpipe/metadata.py
joezuntz: Done.

txpipe/auxiliary_maps.py
joezuntz: Thanks, good point - I've added a standalone variant of the aux mapper that does this.
I think it's still useful to get the pixel scheme from the source maps in the default case, just to avoid risking mismatching resolutions, but the option is there now.

examples/config/cosmodc2_config.yml
joezuntz: See discussion below

txpipe/source_selector.py
joezuntz: Right, so now we examine the file itself to check what kind of data it contains, or whether it has a metadata entry saying what it is.  That way you don't have to keep manually specifying.

examples/laptop_complete_pipeline.yml
joezuntz: good idea - I've done this in the main YAML file for now.

PR #106, opened 2020-06-30
Input pz option

ellongley:


joezuntz:
I've made some changes in https://github.com/LSSTDESC/TXPipe/pull/108 that will impact this, so I'll merge that one first and then fix it up with those.

For that one I'm first waiting on a bug fix being merged in CCL.

PR #109, opened 2020-07-08
Use black -S to tidy up code formatting throughout

joezuntz:
This makes no functional changes whatsoever - it just runs the "[black](https://pypi.org/project/black/)" code formatter on the txpipe and bin directories.

I use the `-S` option to aovid converting all the single quotes to double quotes, since that seemed a bit pointless.

Black includes a check that the code it runs on does not change functionally in any way.


joezuntz:
Will re-open this later.

PR #113, opened 2020-07-20
Adding weights and tests to stats calculators

joezuntz:
This update adds weights to the parallel stats calculators, and rewrite the algorithm used to do so.
It deals with one other issue that the test pulls up, and adds to the sparse array code to do so.

It also adds a test suite for the stats codes, using mock mpi.


joezuntz:
As I've moved this stuff to its own repository, https://github.com/LSSTDESC/parallel_statistics I will close this PR and just re-do it to import the new classes from there.

PR #115, opened 2020-07-23
Adds support for buzzard as input mock catalog

EiffL:
This small PR simply adds the option to create some mock metacal input data out of Buzzard

NOTE: That I couldn't excatly manage to run it for more than one healpix pixel, I'm not sure how @joezuntz does it for CosmoDC2. 

Also I've had to comment out this line:
https://github.com/LSSTDESC/TXPipe/blob/1023b100176205c01ae41e9acb7fcfd9c4883d18/txpipe/input_cats.py#L342
It was failing and I couldn't understand why it was there....

### DIFF
--- a/txpipe/input_cats.py
+++ b/txpipe/input_cats.py
@@ -71,11 +71,18 @@ class TXCosmoDC2Mock(PipelineStage):
             # Load the input catalog (this is lazy)
             # For testing we may want to cut down to a smaller number of pixels.
             # This is separate from the split by processor later on
-            all_healpix_pixels = GCRCatalogs.get_available_catalogs()[cat_name]['healpix_pixels']
+            if 'cosmoDC2' in cat_name:
+                all_healpix_pixels = GCRCatalogs.get_available_catalogs()[cat_name]['healpix_pixels']
+            elif 'buzzard' in cat_name:
+                all_healpix_pixels = GCRCatalogs.load_catalog(cat_name).healpix_pixels
+            else:
+                raise NotImplementedError
+
             max_npix = self.config['max_npix']
             if max_npix != 99999999999999:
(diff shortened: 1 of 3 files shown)


joezuntz:
Does it crash or just take too long on the full catalog? The main sample did take ages to process.

EiffL:
Right, so, I could only use as many processes as I had pixels, I don't know if that's normal. I was trying to run with 6 pixels which I thought would end up being close to the DC2 image area, only managed to run it with 6 processes. And. I was getting a failure on the assert above, so suspicious. And yeah, for 6 pixels, it did take longer than my interactive allocation of about 1 hour

In the end I downgraded to just 1 pixel, and no multiprocessing, and it went through ok.

I guess I'm not sure what configuration of processes/nodes you would recommend using here. I can try to rerun it with better settings.

joezuntz:
I usually ran it in serial, but I think that was because i hadn't discovered `stripe_large` - run that on the directory your output files will be in and it should be much faster.

EiffL:
ok, well, I just regenerated a version with 4 pixels, that's probably gonna be enough for the challenge

txpipe/input_cats.py
```diff
@@ -332,7 +339,7 @@ def write_output(self, start, target_size, photo_cols, metacal_cols, photo_file,
         n = len(photo_data['id'])
         end = min(start + n, target_size)
 
-        assert photo_data['id'].min()>0
```
joezuntz, line 335: I think in CosmoDC2 negative IDs indicated something bad, but that's clearly not important here.

joezuntz:
The error here is some weird issue with Travis failing to download stuff, it's not to do with this PR, so I'll merge.

PR #117, opened 2020-07-25
No shape noise

jpratmarti:
Adding an option in the Covariance stage to set sigma_e to zero when running without shape noise, to get a consistent gaussian covariance.

### DIFF
--- a/examples/config/cosmodc2_config.yml
+++ b/examples/config/cosmodc2_config.yml
@@ -114,7 +114,7 @@ TXTwoPoint:
     max_sep: 250
     nbins: 20
     verbose: 0
-    use_true_shear: True
+    use_true_shear: False
 
 TXTwoPointLensCat:
     binslop: 0.0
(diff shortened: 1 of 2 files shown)


examples/config/cosmodc2_config.yml
```diff
@@ -136,7 +136,9 @@ TXTwoPointFourier:
     bandwidth: 200
     apodization_size: 0.0
     cache_dir: ./cache
-
+    
+TXRealGaussianCovariance:
```
joezuntz, line 140: We probably don't want to set this for the default pipeline config - you could make a second config file if you like?

jpratmarti: Sure makes sense. Will do now. I will just set it to False and whoever wants to run without shape noise can set it to True.

jpratmarti: Done.

joezuntz:
Thanks - just one question - does everything work out okay when you set sigma_e = 0 - there are no errors or anything when you invert the matrix?  Just wanted to check, then will merge.

jpratmarti:
Yes, I was able to compare the no shape noise Gaussian covariance with the JK one and it's not perfect but not crazy either. Thanks!

PR #118, opened 2020-07-29
Switch to the new parallel_statistics library

joezuntz:
This removes txpipe/utils/stats.py and its associated tests and uses the library I span off: https://github.com/LSSTDESC/parallel_statistics


joezuntz:
Since this is just moving bits around and the tests pass I will merge this now.

PR #119, opened 2020-07-29
Weighted calibration

joezuntz:
This PR:
- uses the weights in computing the calibration factors for diagnostics
- adds example lensfit config files
- calibrates the shear maps
- uses lens weights to create density maps - lens weight maps are now returned by the basic Mapper class

Apologies for the mix of things, but they tend to interact in a few ways


txpipe/test/test_cal.py
```diff
@@ -0,0 +1,79 @@
...
+        "mcal_g2_2p": 2*np.array([-0.7, -0.6, -0.4, -0.3, 0.7, 0.6, 0.4, 0.3]) + 0.5*delta_gamma,
+        "mcal_g2_2m": 2*np.array([-0.7, -0.6, -0.4, -0.3, 0.7, 0.6, 0.4, 0.3]) - 0.5*delta_gamma,
+        "weight": np.array([1, 1, 1, 1, 1, 1, 1, 1]),
+    }
+    b1.add_data(data)
+
+    mu, g1, g2, sigma1, sigma2 = b1.collect()
+    assert np.allclose(mu, [-0.5, 0.5])
+    assert np.allclose(g1, [-0.5, 0.5])
+    assert np.allclose(g2, [-1.0, 1.0])
```
ellongley, line 35: Hey Joe, taking a look at this. Is this meant to be one value or std in the [-1., 0., 1.] bins? Where is [-0.2, -0.1, 0.1, 0.2] coming from?

txpipe/test/test_cal.py
```diff
@@ -0,0 +1,79 @@
...
+    expected_sigma1 = np.std([-0.2, -0.1, 0.1, 0.2]) / np.sqrt(4)
```
ellongley, line 36: Hey Joe, taking a look at this. Is this meant to be one value or std in the [-1., 0., 1.] bins? Where is [-0.2, -0.1, 0.1, 0.2] coming from?

joezuntz: To make the test slightly less trivial I was using `var([0.7, 0.6, 0.4, 0.3]) == var([-0.2, -0.1, 0.1, 0.2])`, since we just shift all the numbers off.  The std dev should be the same for the two pixels as one is just the negative of the other.  I've added some explanatory comments.

PR #120, opened 2020-07-30
Generate convergence maps with WLMassMap

joezuntz:
This adds and uses one file with two new pipeline stages in:
- TXConvergenceMaps, which calls WLMassMap to make Kaiser-Squires convergence maps
- TXConvergenceMapPlots, which plots those maps

I have commented them out from running by default in the laptop_pipeline because you need WLMassMap to run them.
I plan to move to a submodule model shortly, where we would include that package with the repo among others, meaning we could enable it by default.


joezuntz:
Thanks!

PR #127, opened 2020-08-17
Add a script to print commands to delete downstream outputs

joezuntz:
This script is designed to help out re-running pipelines when one stage in it and all the subsequent ones needs re-running.  It prints out the delete commands.


chihway:
This looks good and definitely a useful function! I wonder if it should be automatically executed whenever a stage fails? Like is there any reason to keep the output from stages downstream to a failed stage?

joezuntz:
So if a stage fails then the output shouldn't be generated in the first place, because of one of my favourite ceci features: all output files are given the prefix `inprogress_` until the stage is complete, and only then moved to their final location.

So it's only if the stage fails but you later find out it was wrong that this is useful.

PR #130, opened 2020-08-21
Added Heymans et al. (2012) n_eff

ellongley:


cheymans:
@ellongley - there is a new and improved neff estimator now that also includes the impact of the Responsivity (i.e m or R depending on your notation).   
You can find that in Appendix C3 of Joachimi et al 2020: https://arxiv.org/abs/2007.01844
If the responsivity corrections are significant, the inclusion of these terms aren't negligible and feed through to the covariance matrix.   Hope that is useful.

ellongley:
Great, that's very helpful, thank you @cheymans !

Issue #98, opened 2020-06-15
Use observing condition maps in the two point stages


fjaviersanchez:
@ellison-j and @zdu863 are working on using externally-provided observing condition maps to deproject in the `TwoPointFourier` stage. Using this issue to track progress.

fjaviersanchez:
A question for @joezuntz, @damonge, and @anicola: how would we want to add these in real space? Using the method here: https://arxiv.org/pdf/1507.05360.pdf or do we want something "fancier"?

damonge:
That's the standard, I think. In Fourier space we can also quantify the bias to the power spectrum caused by deprojecting the systematics, which is still lacking in real space (this is something that Erika Wagoner and Eduardo Rozo were working on though)

damonge:
@fjaviersanchez should the sky mask be provided together with these? Or is this something that TXPipe currently generates itself? I think we discussed this in Arizona, but I forget what we decided

fjaviersanchez:
Currently, the mask is created within TXPipe using the depth map generated using the HSC-DR1 method. Maybe we should open another issue to add the ability to read an externally-provided mask/depth map?

damonge:
I think we should, because this is not gonna have any information about actual survey geometry (e.g. pixel fractions that have actually been observed), right?

damonge:
And then the product of this issue and the issue you propose should be eventually upgraded to generate systematic maps and mask ready for 2-point calculation from the supreme inputs (but that can be done later)

fjaviersanchez:
Agreed! I think that for this issue we just want to add something that points to a place where the maps live (repo/directory) and these are passed as healpix maps or flat-sky maps to `NaMaster`. The former can easily be done by `supréme` (or by Zilong's tools). The latter would require some fiddling (essentially remapping `healsparse` to flat sky maps) but in any case, this is something that goes beyond this repo/issue.

damonge:
I agree.
I think flat-sky stuff is gonna be phased out anyway, so I wouldn't spend cycles on it.

ellison-j:
Sounds good.
At the moment, I am converting our healsparse maps to Healpix format, writing to a fits file then reading into TXPipe. This is because healsparse is not available in Joe's shifter image (`shifter --image docker:joezuntz/txpipe bash`) and `pip install` does not work in the shifter environment. @joezuntz is it possible to include healspase in the environment?

ellison-j:
[...]

PR #132, opened 2020-08-25
Systematics deprojection

ellison-j:
Added systematics deprojection for galaxy number counts in TXTwoPointFourier.
See https://github.com/LSSTDESC/TXPipe/issues/98.

### DIFF
--- a/txpipe/twopoint_fourier.py
+++ b/txpipe/twopoint_fourier.py
@@ -64,6 +64,9 @@ class TXTwoPointFourier(PipelineStage):
         "flip_g1": False,
         "flip_g2": False,
         "cache_dir": '',
+        "syst_nc": False,
+        "syst_wl": False,
+        "systmaps_nc_dir": '',
     }
 
     def run(self):
@@ -152,7 +155,6 @@ class TXTwoPointFourier(PipelineStage):
     def load_maps(self):
         import pymaster as nmt
         import healpy
-
         # Load the maps from their files.
         # First the mask
         with self.open_input('mask', wrapper=True) as f:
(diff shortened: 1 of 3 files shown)


txpipe/twopoint_fourier.py
```diff
@@ -64,6 +64,9 @@ class TXTwoPointFourier(PipelineStage):
         "flip_g1": False,
         "flip_g2": False,
         "cache_dir": '',
+        "syst_nc": False,
```
joezuntz, line 67: Could you give these a slightly more descriptive name, e.g. `deproject_clustering_sys`?  And also remove the lensing option for now as it doesn't do anything yet.

txpipe/twopoint_fourier.py
```diff
@@ -221,11 +223,85 @@ def load_maps(self):
...
+        # Load HEALPix systematics maps
+        syst_nc = self.config['syst_nc']
+        syst_wl = self.config['syst_wl']
+        if syst_nc:  
+            print('Deprojecting systematics maps for number counts')
+            n_systmaps = 0
+            s_maps = []
+            systmaps_dir = self.config['systmaps_nc_dir']
+            systmaps_path = pathlib.Path(systmaps_dir)
+            try:
```
joezuntz, line 236: should this try statement be inside the loop so that if one map doesn't work the others can still deproject?

txpipe/twopoint_fourier.py
```diff
@@ -221,11 +223,85 @@ def load_maps(self):
...
+                for systmap in systmaps_path.iterdir():
+                    if systmap.is_file():
+                        systmap_file = str(systmap)
```
joezuntz, line 239: maybe also check if the filename ends with .fits or .hs here and only use it if so?

txpipe/twopoint_fourier.py
```diff
@@ -221,11 +223,85 @@ def load_maps(self):
...
+                        dec = -30.
+                        theta = 0.5 * np.pi - np.deg2rad(dec)
+                        phi = np.deg2rad(ra)
+                        nside = healpy.pixelfunc.get_nside(syst_map)
+                        ipix = healpy.ang2pix(nside, theta, phi)
+                        print('Syst map: value at ra,dec = 55,-30: ', syst_map[ipix])
+
+                        # normalize map for Namaster
+                        # set pixel values to value/mean - 1
+                        syst_map_mask = syst_map != healpy.UNSEEN
```
joezuntz, line 254: are there typically any nans in these maps to worry about?

txpipe/twopoint_fourier.py
```diff
@@ -221,11 +223,85 @@ def load_maps(self):
...
+                        mean = np.mean(syst_map[syst_map_mask]) # gives mean of all pixels with mask applied
+                        if mean != 0:
+                            print('Syst map: mean value = ', mean)
+                            syst_map[~syst_map_mask] = 0 # sets unmasked pixels to zero 
+                            syst_map = syst_map / mean - 1
+                            print('Syst map', systmap_file, 'normalized value at ra,dec = 55,-30: ', syst_map[ipix])
+                        
+                        s_maps.append(syst_map)
```
joezuntz, line 262: could you also set something like `self.config[f'clustering_deproject_{n_systmaps}'] = systmap_file` so it gets recorded in the output file provenance?

txpipe/twopoint_fourier.py
```diff
@@ -221,11 +223,85 @@ def load_maps(self):
...
+                raise
+                
+            print('Number of systematics maps read: ', n_systmaps)
+            # We assume all systematics maps have the same nside
+            nside = healpy.pixelfunc.get_nside(syst_map)
+            npix = healpy.nside2npix(nside)
+            # needed for NaMaster:
+            s_maps_nc = np.array(s_maps).reshape([n_systmaps, 1, npix])
+
+# wl not implemented yet
```
joezuntz, line 275: A good question!  Maybe look for pairs of maps?
We could also read a text file of paths to deproject, and look for pairs for shear.  That can wait, though.

joezuntz (changes requested):

txpipe/twopoint_fourier.py
ellison-j: ok

txpipe/twopoint_fourier.py
ellison-j: This was here to catch the problem of no map files in the given directory. If one of the maps is a .fits file that is not a proper healpix maps file `healpy.read_map` will give an error and the job will stop. So I am not sure about moving it.

txpipe/twopoint_fourier.py
ellison-j: I'll check that it is a .fits file since at present only HEALPix is implemented (we can add healsparse later when TXPipe supports it).

txpipe/twopoint_fourier.py
ellison-j: There should not be any NaN's in the HELAPix map as far as I know.

txpipe/twopoint_fourier.py
ellison-j: OK, but I don't understand how the provenance works. Do I just add that line and let the provenance code do the rest or do I have to do something more?

txpipe/twopoint_fourier.py
ellison-j: Yes, we could use ellipticity e1, e2. But will leave lensing systematics out for now.

ellison-j:
Hi @joezuntz I have implemented your suggestions and updated the branch. I am now rejecting any map files that are not valid and using all the others, rather than halting the job if one file is bad.  This is fine I think, but the user needs to check the log file to make sure all maps provided were used.

I also changed the yml file to use only 16 threads since with 64 threads I was running out of memory with 4 maps.

Let me know if you have any more comments.

txpipe/twopoint_fourier.py
joezuntz: Yep, that's right - everything in the `self.config` dictionary gets put in the output file metadata when it gets opened, automatically.

PR #136, opened 2020-09-03
Change how YAML is saved/loaded

joezuntz:
- Adds and uses a FiducialCosmology data type subclass
- Adds helper methods including metadata to YamlFile
- Removes some stray commas in metadata that were making things get saved as tuples

### DIFF
--- a/txpipe/data_types/base.py
+++ b/txpipe/data_types/base.py
@@ -305,13 +305,7 @@ class YamlFile(DataFile):
                 raise ValueError(f"Unknown value {yaml_load} of load_mode. "
                                   "Should be 'safe', 'full', or 'unsafe'")
             # get provenance
-            prov = self.content.pop('provenance', {})
-            self.provenance = {
-                'uuid':     prov.get('uuid', "UNKNOWN"),
-                'creation': prov.get('creation', "UNKNOWN"),
-                'domain':   prov.get('domain', "UNKNOWN"),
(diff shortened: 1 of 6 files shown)


txpipe/data_types/base.py
```diff
@@ -284,9 +284,57 @@ class TextFile(DataFile):
 class YamlFile(DataFile):
     """
     A data file in yaml format.
+    The top-level object in TXPipe YAML
+    files should always be a dictionary.
     """
     suffix = 'yml'
 
+    def __init__(self, path, mode, extra_provenance=None, validate=True, load_mode='full'):
```
empEvil, line 292: `mode` is not an optional argument, but I don't see it given any of the places you changed, am I missing something?

empEvil (commented):
This is the only thing I have questions about.

txpipe/data_types/base.py
joezuntz: when you call `open_input` or `open_output` it automatically selects either `r` or `w` for the mode accordingly.  It can be a bit annoying, though, maybe we should switch to the default being `r`?

empEvil: no I think it is fine then, that just didn't show up in the files changed, so I was unclear. I think that is all good then.

empEvil (approved):

joezuntz:
Thanks!

PR #138, opened 2020-09-14
Add 2D (non-tomographic) lens and calibrated source bins

joezuntz:
This change adds non-tomographic source and lens bins, for example for correlating with systematics.

It:
- adds the counts to the number density stats (simple)
- saves them in the lens selector (simple)
- (added) add 2d counts in redmagic ingest
- makes and calibrates the 2D source map (medium)
- adds a new 2D calibrator for the source selector, reorganizing the selection to avoid copy-paste selections (fiddly)
- reorganizes metadata output slightly (bit fiddly)

### DIFF
--- a/txpipe/maps.py
+++ b/txpipe/maps.py
@@ -172,14 +172,14 @@ class TXSourceMaps(TXBaseMaps):
                 cal = {i:R[i] for i in range(nbin_source)}
                 cal['2D'] = f['/metacal_response/R_total_2d'][:]
             elif shear_catalog_type == "lensfit":
-                R = f['/metacal_response/R_mean'][:]
-                K = f['/metacal_response/K'][:]
-                c = f['/metacal_response/C'][:]
+                R = f['/response/R_mean'][:]
+                K = f['/response/K'][:]
+                c = f['/response/C'][:]
                 cal = {i: (R[i], K[i], c[i]) for i in range(nbin_source)}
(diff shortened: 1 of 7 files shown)


txpipe/maps.py
```diff
@@ -178,6 +162,29 @@ def prepare_mappers(self, pixel_scheme):
...
+    def get_calibrators(self):
+        shear_catalog_type = read_shear_catalog_type(self)
+        with self.open_input("shear_tomography_catalog") as f:
+            nbin_source = f["tomography"].attrs["nbin_source"]
+
+            if shear_catalog_type == "metacal":
+                R = f['/metacal_response/R_total'][:] # nbin x 2 x 2
+                cal = {i:R[i] for i in range(nbin_source)}
+                cal['2D'] = f['/metacal_response/R_total_2d'][:]
+            elif shear_catalog_type == "lensfit":
```
jpratmarti, line 174: Why does the lensfit option also have responses to calibrate? Isn't the difference between the first and second option to use the full matrix response or just the mean response?

jpratmarti (approved):

txpipe/maps.py
joezuntz: I don't know much about the lensfit cal - just adapting what @ellongley did but for the 2D case. There's a function in `calibration_tools.py` that handles it all

PR #139, opened 2020-09-15
Gamma x

empEvil:
This pull request ensures we save the GammaX correlations from TreeCorr.

- It saves the GammaX to a separate sacc file.
- It uses the shot noise from TreeCorr to calculate the error and the covariance.
- When plotting with the stage TwoPointPlots, it now also plots the GammaX.

### DIFF
--- a/examples/laptop_pipeline.yml
+++ b/examples/laptop_pipeline.yml
@@ -32,7 +32,6 @@ stages:
     - name: TXPSFDiagnostics     # Compute and plots other PSF diagnostics
     - name: TXBrighterFatterPlot # Make plots tracking the brighter-fatter effect
     - name: TXPhotozPlots        # Plot the bin n(z)
-    #- name: TXTwoPGXP            # Plot the gammaX correlation
 
     # These two require the WLMassMap repository on the path - see below
     # - name: TXConvergenceMaps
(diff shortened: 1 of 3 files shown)


examples/laptop_pipeline.yml
```diff
@@ -32,6 +32,7 @@ stages:
     - name: TXPSFDiagnostics     # Compute and plots other PSF diagnostics
     - name: TXBrighterFatterPlot # Make plots tracking the brighter-fatter effect
     - name: TXPhotozPlots        # Plot the bin n(z)
+    #- name: TXTwoPGXP            # Plot the gammaX correlation
```
jpratmarti, line 35: Is this supposed to be commented or should we comment it out?

empEvil: Was a left over from trying to get plotting to work

jpratmarti (approved):

examples/laptop_pipeline.yml
jpratmarti: Actually @empEvil I just seen that you have some conflicts with master, so you should probably resolve this before merging (sorry I approved too fast! :) )

empEvil:
I think I fixed the last outstanding conflicts with master

jpratmarti (approved):

joezuntz (changes requested):
This looks great - I've requested one very minor change.

PR #140, opened 2020-09-17
Submodules

joezuntz:
Using a new ceci v1.1 feature, include WLMassMap, FlexZPipe, and TJPCov as submodules inside the main repo.

Updates the pipeline and config files to use these new paths, and since the convergence map is now automatic we run that always.


joezuntz:
Since this doesn't change any code, only configuration and README, I'm giving myself permission to merge it.

PR #141, opened 2020-09-17
Compute correlations between data and systematic maps

joezuntz:
Adds TXMapCorrelations class that makes plots correlating g1, g2, convergence, and number counts with Supreme property maps.

Three supporting changes are included too:

- a 2D convergence map
- a FileCollection data type for the outputs
- a list_maps method to MapsFile (no longer used in this PR but useful interactively)


txpipe/map_tests.py
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    def read_healsparse(self, map_path, nside):
+        import healsparse
+        import healpy
+
+        # Convert to correct res healpix map
+        m = healsparse.HealSparseMap.read(map_path)
+        m = m.generate_healpix_map(nside=nside)
+        m = healpy.ud_grade(m, nside, order_in="nest", order_out="ring")
```
chihway, line 33: I think the recommendation is not to use healpy's `ud_grade` function as it's not doing things 100% correctly. Better to do map2alm and alm2map (setting the nside here)

txpipe/map_tests.py
```diff
@@ -0,0 +1,162 @@
...
+            source_g1 = map_file.read_map("g1_2D")
+            source_g2 = map_file.read_map("g2_2D")
+
+        with self.open_input("convergence_maps", wrapper=True) as map_file:
+            kappa = map_file.read_map("kappa_E_2D")
+
+        with self.open_input("mask", wrapper=True) as map_file:
+            mask = map_file.read_map("mask")
+
+        if not (kappa.size == ngal.size):
```
chihway, line 59: do we also need to check the mask size?

examples/config/laptop_config.yml
```diff
@@ -190,3 +190,6 @@ TXTwoPointFourier:
 TXSimpleMask:
     depth_cut : 23.5
     bright_object_max: 10.0
+
+TXMapCorrelations:
+    supreme_path_root: data/example/inputs/supreme
```
chihway, line 195: can we add "outlier fraction" and "nbin" here just to make it clear you can set it?

txpipe/map_tests.py
```diff
@@ -0,0 +1,162 @@
...
+    def correlate(self, sys_map, data_map, mask):
+        import scipy.stats
+        import healpy
+
+        N = self.config["nbin"]
+        f = 0.5 * self.config["outlier_fraction"]
+
+        # clean the data
+        finite = (
+            np.isfinite(sys_map + data_map)
```
chihway, line 108: why don't do them individually?

txpipe/map_tests.py
```diff
@@ -0,0 +1,162 @@
...
+        counts = np.bincount(bins)
+
+        # and the mean of the systematic value
+        # itself, and of the y and y^2 values
+        x = np.bincount(bins, weights=sys_map) / counts
+        y = np.bincount(bins, weights=data_map) / counts
+        y2 = np.bincount(bins, weights=data_map ** 2) / counts
+
+        # the var on the mean = var / count
+        yerr = np.sqrt((y2 - y**2) / counts)
```
chihway, line 140: do you need to divide by counts?

chihway: ohhh, I guess you're plotting the error on the mean here? in that case nvm

txpipe/__init__.py
```diff
@@ -30,3 +30,4 @@
 from .masks import TXSimpleMask
 from .metadata import TXTracerMetadata
 from .convergence import TXConvergenceMaps
+from .map_tests import TXMapCorrelations
```
chihway, line 33: wonder if the map_tests name can be a little more descriptive, I think map_correlation could be better?

txpipe/data_types/types.py
```diff
@@ -1,4 +1,5 @@
-from .base import FitsFile, HDFFile, DataFile, YamlFile, TextFile, Directory, PNGFile
+from .base import FitsFile, HDFFile, DataFile, YamlFile
+from .base import TextFile, Directory, PNGFile, FileCollection
```
chihway, line 2: not a big deal, but FileCollection is not used here (I was trying to look at how it gets used)

chihway:
Ok @joezuntz I put in some random comments, nothing major. I wasn't able to actually run it though since nersc was down...

txpipe/map_tests.py
joezuntz: I didn't know that - interesting!  Although for this case the nside for the map will not change, so it's just re-indexing the pixels from ring to nest (because we generated the map at the correct nside in the first place), which I assume should be fine?  Having said that, we are using healsparse's generate_healpix map to go from the input to the output nside, and I have no idea how that works - will check.

examples/config/laptop_config.yml
joezuntz: done.

txpipe/map_tests.py
joezuntz: yep, error on the mean.  which I think is the right thing to plot.

txpipe/data_types/types.py
joezuntz: I've tidied up the whole thing.

txpipe/map_tests.py
joezuntz: Done!

txpipe/map_tests.py
joezuntz: changed.

txpipe/__init__.py
joezuntz: changed.

txpipe/map_tests.py
joezuntz: By default healsparse just takes the mean of hit pixels in the high-res map, which I think is the right thing to do here.

joezuntz: I've added a comment noting that the resolution doesn't change.

joezuntz:
Merging since I've addressed the notes from @chihway and the tests pass.

PR #143, opened 2020-09-18
Ell binning and plots

jpratmarti:
Adding the following features:
- Possibility to easily change the ell binning. Added the following options in the config file in `twopoint_fourier.py`:
        "ell_min": 100,
        "ell_max": 1500,
        "n_ell": 20,
        "ell_spacing": 'log'
- Switched the plotting script to load `summary_statistics_fourier.sacc` instead of `twopoint_data_fourier.sacc` to include errorbars in the plots (the covariance is now in the `summary_statistics_fourier.sacc` file).

### DIFF
--- a/txpipe/twopoint_fourier.py
+++ b/txpipe/twopoint_fourier.py
@@ -454,7 +454,7 @@ class TXTwoPointFourier(PipelineStage):
 
         # Creating the ell binning from the edges using this Namaster constructor.
         edges = np.unique(np.geomspace(self.config['ell_min'], self.config['ell_max'], self.config['n_ell']).astype(int))
-        ell_bins = MyNmtBin.from_edges(edges_int[:-1], edges_int[1:], is_Dell=False)
+        ell_bins = MyNmtBin.from_edges(edges[:-1], edges[1:], is_Dell=False)
         return ell_bins
 
 
--- a/txpipe/twopoint_fourier.py
+++ b/txpipe/twopoint_fourier.py
@@ -60,7 +60,6 @@ class TXTwoPointFourier(PipelineStage):
 
     config_options = {
         "mask_threshold": 0.0,
-        "bandwidth": 0,
         "flip_g1": False,
         "flip_g2": False,
         "cache_dir": '',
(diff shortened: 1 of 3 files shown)


txpipe/twopoint_fourier.py
```diff
@@ -439,11 +446,16 @@ def choose_ell_bins(self, pixel_scheme, f_sky):
...
+        '''
+        
         # The subclass of NmtBin that we use here just adds some
         # helper methods compared to the default NaMaster one.
         # Can feed these back upstream if useful.
-        ell_bins = MyNmtBin(int(pixel_scheme.nside), nlb=nlb)
+
+        # Creating the ell binning from the edges using this Namaster constructor.
+        edges = np.geomspace(self.config['ell_min'], self.config['ell_max'], self.config['n_ell'])
+        edges_int = edges.astype(int)
```
damonge, line 457: The problem with this is that sometimes, after casting to `int`, the result of `geomspace` will have repeated values. What I usually do is:
```
edges = np.unique(np.geomspace(lmin, lmax, n_ell).astype(int))
```

damonge (commented):
Hi @jpratmarti 

looks good, just a tiny comment on repeated ell values.

Also, I see you guys are using the `MyNmtBin` wrapper. Is there anything missing in the standard `NmtBin`s that you'd like to have (e.g. there's already a `from_edges` class method)? If there is, I'd be happy to implement it and then switch to them it in a separate PR.

Also, it seems there are conflicts with master, so you should sort those out ;-)

joezuntz:
@damonge I'll do a PR on namaster with the changes we have here if that would be welcome.
Joe

damonge:
Yes, sure! Thanks @joezuntz

txpipe/twopoint_fourier.py
jpratmarti: Great thanks a lot David for reviewing this. I added your solution to the repeating values, and will now look at the conflicts with master.

damonge (approved):
LGTM!

joezuntz:
Two more quick things - could you remove the old `bandwidth` option that no longer does anything, and either implement the log-spacing option or remove it?

PR #146, opened 2020-10-24
No noise fourier

jpratmarti:
Adding options when need to obtain a Fourier space measurement on cosmodc2 without shape noise and corresponding covariance.

### DIFF
--- a/examples/config/cosmodc2_config.yml
+++ b/examples/config/cosmodc2_config.yml
@@ -26,18 +26,18 @@ TXSourceTrueNumberDensity:
 TXMainMaps:
     chunk_rows:  100000
     pixelization: healpix
-    nside: 2048
+    nside: 512
     sparse: True
 
 TXSourceMaps:
-    nside: 2048
+    nside: 512
     sparse: True
     chunk_rows:  100000
     pixelization: healpix
     true_shear: False
 
 TXExternalLensMaps:
-    nside: 2048
+    nside: 512
     sparse: True
     chunk_rows:  100000
     pixelization: healpix
--- a/examples/config/cosmodc2_config.yml
+++ b/examples/config/cosmodc2_config.yml
@@ -26,17 +26,18 @@ TXSourceTrueNumberDensity:
 TXMainMaps:
     chunk_rows:  100000
     pixelization: healpix
-    nside: 512
+    nside: 2048
     sparse: True
 
 TXSourceMaps:
-    nside: 512
+    nside: 2048
     sparse: True
     chunk_rows:  100000
     pixelization: healpix
+    true_shear: False
 
 TXExternalLensMaps:
-    nside: 512
+    nside: 2048
     sparse: True
     chunk_rows:  100000
     pixelization: healpix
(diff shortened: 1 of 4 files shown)


txpipe/twopoint_fourier.py
```diff
@@ -540,12 +541,15 @@ def compute_power_spectra(self, pixel_scheme, i, j, k, maps, workspaces, ell_bin
         workspace = workspaces[(i,j,k)]
 
         # Get the coupled noise C_ell values to give to the master algorithm
-        cl_noise = self.compute_noise(i, j, k, ell_bins, maps, workspace)
+        if self.config['true_shear']:
```
joezuntz, line 544: Isn't this the wrong way around?

You could also put it inside the compute_noise function to make the top level look simpler.

txpipe/maps.py
```diff
@@ -213,8 +213,8 @@ def data_iterator(self):
     def accumulate_maps(self, pixel_scheme, data, mappers):
         # rename columns, if needed
         if self.config["true_shear"]:
-            data["g1"] = data["true_g1"]
-            data["g2"] = data["true_g2"]
+            data["g1"] = data[f"true_g1"]
```
joezuntz, line 216: The "f"s here aren't doing anything.

examples/config/cosmodc2_config.yml
```diff
@@ -26,17 +26,18 @@ TXSourceTrueNumberDensity:
 TXMainMaps:
     chunk_rows:  100000
     pixelization: healpix
-    nside: 512
```
joezuntz, line 29: Could we leave these values at their previous ones for the default pipeline?  Could add a second config file maybe for the high res and true shear version.

txpipe/twopoint_fourier.py
```diff
@@ -66,9 +66,10 @@ class TXTwoPointFourier(PipelineStage):
         "deproject_syst_clustering": False,
         "systmaps_clustering_dir": '',
         "ell_min": 100,
-        "ell_max": 1500,
-        "n_ell": 20,
-        "ell_spacing": 'log'
+        "ell_max": 3000,
```
joezuntz, line 69: Can leave these changes in as they don't hurt.

joezuntz (changes requested):

txpipe/twopoint_fourier.py
jpratmarti: Well, if we change nside to 512 again we can't reach that higher ell. What should we use? Should I leave nside 2048 or return to ell_max 1500?

txpipe/twopoint_fourier.py
jpratmarti: Yes, thanks for catching this, I just changed the option at the last moment and forgot to switch the boolean.
I also put the it inside the compute_noise function.

examples/config/cosmodc2_config.yml
joezuntz: actually, can leave these as they are I think, this is the CosmoDC2 file after all which has high number density.

txpipe/twopoint_fourier.py
joezuntz: yeah, good point!  Can we leave it on the original defaults just because they're faster for the laptop tests?

examples/config/cosmodc2_config.yml
joezuntz: I mean, leave your changes in place.

txpipe/twopoint_fourier.py
```diff
@@ -602,7 +603,11 @@ def compute_noise(self, i, j, k, ell_bins, maps, workspace):
         if k == POS_POS:
             mean_noise /= 4
 
-        return mean_noise
+        if self.config['true_shear']:
```
joezuntz, line 606: Can you do this test at the start of the function and avoid the calculation if it's not needed?

PR #154, opened 2021-01-07
Interface with RAIL for PZ calculations

joezuntz:
This uses the RAIL Estimator class for PZ PDF calculations.


txpipe/data_types/base.py
```diff
@@ -444,4 +445,28 @@ def write_provenance(self):
...
-        raise ValueError("Reading existing PNG files is not supported")
\ No newline at end of file
+        raise ValueError("Reading existing PNG files is not supported")
+
+class PickleFile(DataFile):
+    suffix = "pkl"
+
+    @classmethod
+    def open(self, path, mode, **kwargs):
+        return open(path, mode + "b")
```
johannct, line 455: should self.mode be updated?

txpipe/data_types/base.py
```diff
@@ -444,4 +445,28 @@ def write_provenance(self):
...
+        return self.read()
+
+    def write(self, obj):
+        if self.mode != "w":
+            raise ValueError("Tried to write to read-only pickle file")
+        pickle.dump(obj, self.file)
+
+    def read(self):
+        if self.mode != "r":
+            raise ValueError("Tried to read from a write-only pickle file")
```
johannct, line 470: I would add self.mode to the print statement, in order to help users in debugging (this is true in several places, and maybe a ModeError exception is in order)

txpipe/rail_pz.py
```diff
@@ -0,0 +1,190 @@
...
+    This stage, which would normally be run first, uses a training set (e.g. of
+    spectroscopic redshifts) to train and save an Estimator object that a later
+    stage can use to measure redshifts of the survey sample.
+    """
+    name = "PZRailTrain"
+
+    inputs = [
+        ("photoz_training", HDFFile),
+        ("photoz_testing", HDFFile),
+    ]
```
johannct, line 22: is there a way to be more agnostic at this stage, about what format the inputs will actually be in?

joezuntz: That's a good idea - I've changed them to use the generic `DataFile` type, which can be anything.

txpipe/rail_pz.py
```diff
@@ -0,0 +1,190 @@
...
+
+        # Additional confguration specific to this algorithm
+        # can be included in the config
+        run_dict = {
+            "run_params": self.config.copy(),
+        }
+
+        # create an instance of the specific estimation
+        # method
+        cls = Estimator._find_subclass(self.config["class_name"])
```
johannct, line 57: looks like the underscore is not compatible with such a use, and not warranted, as this usage seems reasonable. I wonder if we can/should add classmethod __call__ to Estimator, that would hide the exact same call.

txpipe/rail_pz.py
```diff
@@ -0,0 +1,190 @@
...
+        output = self.setup_output_file(estimator)
+
+        # Create the iterator the reads chunks of photometry
+        # The method we use here automatically splits up data when we run in parallel
+        cols = [f"mag_{b}" for b in "ugrizy"] + [f"mag_err_{b}" for b in "ugrizy"]
+        chunk_rows = self.config["chunk_rows"]
+        it = self.iterate_hdf("photometry_catalog", "photometry", cols, chunk_rows)
+
+        # Loop through the chunks of data
+        for s, e, data in it:
```
johannct, line 126: implement pooling here? especially if parallel write is available, but parallel run can always be done and gathered for sequential write ex post. Ideally all this should be hidden in the estimator codebase

txpipe/rail_pz.py
```diff
@@ -0,0 +1,190 @@
...
+        # the suffix _lsst, which we add here
+        for band in "ugrizy":
+            data[f"mag_{band}_lsst"] = data[f"mag_{band}"]
+            data[f"mag_err_{band}_lsst"] = data[f"mag_err_{band}"]
+
+    def setup_output_file(self, estimator):
+        # Briefly check the size of the catalog so we know how much
+        # space to reserve in the output
+        with self.open_input("photometry_catalog") as f:
+            nobj = f["photometry/ra"].size
```
johannct, line 147: what format needs `f` be for `"photometry/ra"` to be legit?

txpipe/data_types/base.py
joezuntz: I've changed this to `io.UnsupportedError`, which is what you get if you try to do the equivalent thing with an ordinary python file object, and added a print of self.mode.

I've also added a check when opening the file in the first place that the argument is either "r" or "w".

txpipe/data_types/base.py
joezuntz: I'd like to keep this as "r" or "w" throughout, as below.  This is consistent with what we do with FITS files.

txpipe/rail_pz.py
joezuntz: I've asked about this here: https://github.com/LSSTDESC/RAIL/issues/45#issuecomment-756201258

txpipe/rail_pz.py
joezuntz: It's already automatically parallel here :-)

The `iterate_hdf` method knows about parallelism and will split up the input data among MPI processes, so the iterator only returns data chunks (including the `s` and `e` indices that are assigned to this process).  And because we opened the file with `parallel=True` in `setup_output_file` then parallel write is automatically enabled.

txpipe/rail_pz.py
joezuntz: HDF5.   Which it always is here - this photometry input file here is the one we use within TXPipe, which we first extract from GCR to disc in an earlier stage (in `input_cats.py`).

txpipe/rail_pz.py
johannct: why is there a for loop here?

joezuntz: The `iterate_hdf` method is a generator, so it `yield`s chunks of data one by one when it's used in a for loop.

We split up both by process and by chunk, so e.g. if you have two processors and chunk size 10 with 100 data points then process 1 will be given chunks 0-9, 20-29, 40-49, 60-69, 80-89, but not all at once - it is given the one by one to minimize memory usage.

txpipe/rail_pz.py
johannct: ok I see, we stopped short from defining a write interface for pz products, but we are probably going to have to do it, to ensure consistency..... so this will need to be revised anyway

txpipe/rail_pz.py
johannct: got it.

txpipe/rail_pz.py
joezuntz: I have a parallel issue to switch to using `qp` for data storage.  Once I have this bit worked out I'll look into that.

sschmidt23: There is already a RAIL branch with a quick and dirty switch to qp storage. 
https://github.com/LSSTDESC/RAIL/tree/issue/33/qp-output
However, as mentioned in the associated issue 
https://github.com/LSSTDESC/RAIL/issues/33
there is currently an awkward hack where we write to multiple astropy table chunks as an interim step and then join them at the end in an inefficient way (note that RAIL has its own data chunking separate from TXPipe).
If there's a way to write astropy tables with parallel writes I'd be happy to switch this out. Or, we may want to talk about switching qp storage to just straight hdf5 rather than astropy tables, since parallel write is pretty straightforward for that format.

sjoudaki:
@joezuntz As requested, I formally accept this as PZ convenor.

PR #159, opened 2021-01-25
added wl deprojection

ellison-j:
Added deprojection for shear. Uses all maps in specified directory. If map file name contains "_e1","e_2" assumes it is a pair. If not, assumes it is a scalar map.

### DIFF
--- a/txpipe/twopoint_fourier.py
+++ b/txpipe/twopoint_fourier.py
@@ -222,7 +222,7 @@ class TXTwoPointFourier(PipelineStage):
                 w = np.where(g2!=healpy.UNSEEN)
                 g2[w]*=-1
                 
-
+                
         # Load HEALPix systematics maps for clustering deprojection
         deproject_syst_clustering = self.config['deproject_syst_clustering']
         if deproject_syst_clustering:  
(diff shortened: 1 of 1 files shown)


txpipe/twopoint_fourier.py
```diff
@@ -231,37 +234,15 @@ def load_maps(self):
             for systmap in systmaps_path.iterdir():
                 try:
                     if systmap.is_file():
-                        if pathlib.Path(systmap).suffix not in [".fits", ".hs"]:
-                            print('Warning: Problem reading systematics map file', systmap)
-                            print('Not a HEALPix .fits or a healsparse .hs file.')
-                            warnings.warn("Systematics map file must be a HEALPix .fits file or a healsparse .hs file.")
+                        if pathlib.Path(systmap).suffix != ".fits":
```
fjaviersanchez, line 237: I'd suggest to remove this since the suffix can be anything, as long as the file is in fits format, also healpy can read `.fits.gz` files natively. Maybe it's better to do:

```
try:
    if systmap.is_file():
        syst_map = healpy.read_map(systmap_file, verbose=False)
    else:
        warnings.warn(f'Map from {systmap} is missing. Skipping...')
except:
    print(f'A problem occurred while reading {systmap}. Skipping...')
    continue
```

txpipe/twopoint_fourier.py
```diff
@@ -289,8 +270,144 @@ def load_maps(self):
...
+            print('Number of scalar systematics maps read: ', n_systmaps_scalar)
+                    
+            # load _e1, _e2 maps
+            n_systmaps = 0
+            n_systmaps_e1 = 0
+            for file in s_maps_e1_names:
+                systmap = pathlib.Path(file)
+                try:
+                    if systmap.is_file():
+                        if pathlib.Path(systmap).suffix != ".fits":
```
fjaviersanchez, line 335: same comment as above for the clustering part.

txpipe/twopoint_fourier.py
```diff
@@ -484,19 +606,19 @@ def select_calculations(self, nbins_source, nbins_lens):
             for j in range(i + 1):
                 calcs.append((i, j, k))
 
-        # For shear-position we use all pairs
-        k = SHEAR_POS
-        for i in range(nbins_source):
-            for j in range(nbins_lens):
-                calcs.append((i, j, k))
+#         # For shear-position we use all pairs
```
fjaviersanchez, line 609: We need to keep these calculations

txpipe/twopoint_fourier.py
```diff
@@ -484,19 +606,19 @@ def select_calculations(self, nbins_source, nbins_lens):
...
-        # We do keep cross-pairs, since even though we may not want to
-        # do parameter estimation with them they are useful diagnostics.
-        k = POS_POS
-        for i in range(nbins_lens):
-            for j in range(i + 1):
-                calcs.append((i, j, k))
+#         # For position-position we omit pairs with j>i.
+#         # We do keep cross-pairs, since even though we may not want to
+#         # do parameter estimation with them they are useful diagnostics.
+#         k = POS_POS
```
fjaviersanchez, line 618: We need to keep these calculations as well

txpipe/twopoint_fourier.py
```diff
@@ -539,15 +661,15 @@ def compute_power_spectra(self, pixel_scheme, i, j, k, maps, workspaces, ell_bin
             field_j = maps['lf'][j]
             results_to_use = [(0, CEE, ), (1, CEB, ), (3, CBB, )]
 
-        elif k == POS_POS:
-            field_i = maps['df'][i]
-            field_j = maps['df'][j]
-            results_to_use = [(0, Cdd)]
+#         elif k == POS_POS:
+#             field_i = maps['df'][i]
```
fjaviersanchez, line 665: Same as above

txpipe/twopoint_fourier.py
```diff
@@ -539,15 +661,15 @@ def compute_power_spectra(self, pixel_scheme, i, j, k, maps, workspaces, ell_bin
...
+#             field_j = maps['df'][j]
+#             results_to_use = [(0, Cdd)]
 
-        elif k == SHEAR_POS:
-            field_i = maps['lf'][i]
-            field_j = maps['df'][j]
-            results_to_use = [(0, CdE), (1, CdB)]
+#         elif k == SHEAR_POS:
+#             field_i = maps['lf'][i]
```
fjaviersanchez, line 670: same, same

fjaviersanchez (changes requested):
Sorry for the delay @ellison-j. Thanks a lot for the changes. I have some minor comments that I think will help later on. I believe that some commented lines maybe sneaked in from your tests as well but we'll need the cross-correlations for the project.

txpipe/twopoint_fourier.py
ellison-j: Thanks Javi. I'll make the changes.

txpipe/twopoint_fourier.py
ellison-j: ok

txpipe/twopoint_fourier.py
ellison-j: oops, yes I'll uncomment this and the others below. I think we should add config-based selection of the calculations later as the deprojection can be slow.

fjaviersanchez (approved):
Changes look good! Thanks!!

PR #163, opened 2021-02-02
Ccl tests

jpratmarti:
Add Theory stage to save CCL theory into sacc files, both for real and fourier space. 
Add code to plot ratios of measurements/theory.
Set halofit as default instead of ccl-emu.

### DIFF
--- a/bin/get_bin_edges_equal_comoving.py
+++ b/bin/get_bin_edges_equal_comoving.py
@@ -1,0 +1,30 @@
+import jax_cosmo
+import jax_cosmo.background
+import numpy as np
+
+def get_z_edges(n_bin, zmax):
+    zgrid = np.linspace(0, zmax, 1000)
+    agrid = 1 / (1 + zgrid)
+    model = jax_cosmo.parameters.Planck15()
+    chi_grid = jax_cosmo.background.radial_comoving_distance(model, agrid)
+    # Compute bin edges that are equally spaced in chi.
+    chi_edges = np.linspace(0, chi_grid[-1], n_bin + 1)
+    z_edges = np.empty(n_bin + 1)
+    z_edges[0] = 0.
+    z_edges[-1] = zmax
+    z_edges[1:-1] = np.interp(chi_edges[1:-1], chi_grid, zgrid)
+    print('n_bin, zmax:', n_bin, zmax)
+    print('z edges:', z_edges)
+
+
+n_bin = 5
+zmax = 2.
+get_z_edges(n_bin, zmax)
+
+n_bin = 8
+zmax = 3.
+get_z_edges(n_bin, zmax)
+
+
+
+
--- a/bin/get_bin_edges_equal_comoving.py
+++ b/bin/get_bin_edges_equal_comoving.py
@@ -13,6 +13,7 @@ def get_z_edges(n_bin, zmax):
     z_edges[0] = 0.
     z_edges[-1] = zmax
     z_edges[1:-1] = np.interp(chi_edges[1:-1], chi_grid, zgrid)
+    print('Cosmology used:', model)
     print('n_bin, zmax:', n_bin, zmax)
     print('z edges:', z_edges)
 
--- a/txpipe/plotting/correlations.py
+++ b/txpipe/plotting/correlations.py
@@ -229,8 +229,6 @@ def make_plot(corr, obs_data, obs_theory, fig=None, xlogscale=True, ratios=False
                 res = obs[(corr, i, j)]
                 if ratios:
                     res_theory = obs_theory[index][(corr, i, j)]
-                #if ratios:
-                #    theta_th, xi_th = theory_data[index][(corr, i, j)]
                 if len(res) == 2:
                     theta, xi = res
                     if ratios:
@@ -249,7 +247,7 @@ def make_plot(corr, obs_data, obs_theory, fig=None, xlogscale=True, ratios=False
                     theta, xi, cov = res
                     err = cov.diagonal()**0.5
                     if ratios:
-                        theta_th, xi_th, cov_th = res_theory
+                        theta_th, xi_th = res_theory
                         a.errorbar(theta, xi/xi_th, err/xi_th, fmt='.', label='TXPipe Data/CCL Theory')
                         a.axhline(y=1, color='k', ls = ':')
                     else:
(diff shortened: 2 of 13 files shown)


txpipe/plotting/correlations.py
```diff
@@ -374,11 +498,21 @@ def make_theory_plot_data(data, cosmo, obs, label, smooth=True, xi=None):
             # Optionally also compute correlation functions
             if xi:
                 theta, *_ = obs[(GAMMA, i, j)]
-                try:
```
joezuntz, line 377: I think the except version is better here.  Looks like it's commented here - do you still get the double free in the newest ccl?

txpipe/twopoint.py
```diff
@@ -6,6 +6,7 @@
 import random
 import collections
 import sys
+import pdb
```
joezuntz, line 9: unused

txpipe/twopoint.py
```diff
@@ -739,6 +740,226 @@ def load_lens_catalog(self, data):
...
+
+        # Real space
+        filename = self.get_input('summary_statistics_real')
+        s = sacc.Sacc.load_fits(filename)
+
+        # TODO: when there is a better Cosmology serialization method
+        # switch to that
+        print("Manually specifying matter_power_spectrum and Neff")
+        cosmo = self.open_input('fiducial_cosmology', wrapper=True).to_ccl(
+        matter_power_spectrum='halofit', Neff=3.04)
```
joezuntz, line 770: indent looks odd

txpipe/twopoint.py
```diff
@@ -739,6 +740,226 @@ def load_lens_catalog(self, data):
...
+        # Fourier space
+        filename = self.get_input('summary_statistics_fourier')
+        s = sacc.Sacc.load_fits(filename)
+
+        s_theory = self.replace_with_theory_fourier(s, cosmo, tracers, nbin_source, nbin_lens)
+        # save the output to Sacc file
+        s_theory.save_fits(self.get_output('twopoint_theory_fourier'), overwrite=True)
+
+        
+    def read_nbin(self, s):
```
joezuntz, line 786: at some point we should just save this in metadata

txpipe/twopoint.py
```diff
@@ -739,6 +740,226 @@ def load_lens_catalog(self, data):
...
+            # Convert to CCL form
+            tracers[name] = pyccl.NumberCountsTracer(cosmo, has_rsd=False, 
+                dndz=(Ti.z, nz), bias=(Ti.z, np.ones_like(Ti.z)))
+            
+        return tracers
+    
+    def replace_with_theory_real(self, s, cosmo):
+
+        import pyccl
+        nbin_source, nbin_lens = self.read_nbin(s)
```
joezuntz, line 837: is this all copied from the theory.py?  might be better to avoid repeating ourselves and have this in one place. There's a principle D.R.Y. - Don't Repeat Yourself.

txpipe/twopoint.py
```diff
@@ -779,8 +1004,11 @@ def run(self):
         # TODO: when there is a better Cosmology serialization method
         # switch to that
         print("Manually specifying matter_power_spectrum and Neff")
+        #cosmo = self.open_input('fiducial_cosmology', wrapper=True).to_ccl(
```
joezuntz, line 1007: remove old lines

txpipe/plotting/correlations.py
```diff
@@ -264,6 +267,117 @@ def make_plot(corr, obs_data, theory_data, fig=None, xlogscale=True):
...
+            if j not in J:
+                f.delaxes(a)
+                continue
+
+            for index, obs in enumerate(obs_data):
+
+                res = obs[(corr, i, j)]
+                theta_th, xi_th = theory_data[index][(corr, i, j)]
+                if len(res) == 2:
+                    theta, xi = res
```
joezuntz, line 352: looks like in this case (when there's no error bars) this still plots xi not xi / xi_th

txpipe/plotting/correlations.py
```diff
@@ -339,13 +456,20 @@ def make_theory_plot_data(data, cosmo, obs, label, smooth=True, xi=None):
         for j in range(i+1):
             print(f"Computing theory lensing-lensing ({i},{j})")
 
+            # uncomment to take ratio in Fourier space to get the exact ells
```
joezuntz, line 459: Is the i==0 and j==0 test here just to do it once?  Any reason not to do it each time in case ell changes?

txpipe/twopoint.py
```diff
@@ -739,6 +740,226 @@ def load_lens_catalog(self, data):
...
+                else:
+                    gt = pyccl.correlation(cosmo, ell, cl, theta/60, corr_type='GL')
+                    
+                # replace data values in the sacc object for the theory ones
+                ind = s.indices('galaxy_shearDensity_xi_t', (f'source_{i}', f'lens_{j}'))
+                for p, q in enumerate(ind):
+                    s.data[q].value = gt[p]
+
+
+                # somehow the exception still gives an error, only the if statement above works.
```
joezuntz, line 895: need to fix this properly not with this band aid.  Is this in the newest CCL?

txpipe/twopoint_fourier.py
```diff
@@ -763,8 +764,13 @@ def run(self):
         s = sacc.Sacc.load_fits(filename)
         nbin_source, nbin_lens = self.read_nbin(s)  
  
-        cosmo = self.open_input('fiducial_cosmology', wrapper=True).to_ccl()
-
+        print("Manually specifying matter_power_spectrum and Neff")
+        #cosmo = self.open_input('fiducial_cosmology', wrapper=True).to_ccl(
```
joezuntz, line 768: more old lines to remove

joezuntz (changes requested):

bin/get_bin_edges_equal_comoving.py
```diff
@@ -0,0 +1,30 @@
+import jax_cosmo
```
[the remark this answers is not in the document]
jpratmarti: Yeah I just wanted to be completely consistent with what they used in the tomo challenge so I would keep it as is for the moment. Maybe we can revisit at a later time.

txpipe/plotting/correlations.py
```diff
@@ -121,7 +121,10 @@ def full_3x2pt_plots(sacc_files, labels,
     for t in types:
         if any(obs[t] for obs in obs_data):
             f = figures.get(t)
-            output_figures[t] = make_plot(t, obs_data, theory_data, fig=f, xlogscale=xlogscale)
+            if not ratios:
+                output_figures[t] = make_plot(t, obs_data, theory_data, fig=f, xlogscale=xlogscale)
+            if ratios:
```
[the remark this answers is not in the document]
jpratmarti: Yes agreed, thank you for the tip.

txpipe/twopoint.py
jpratmarti: yeah agreed. I think the best way would be to load the theory files in the plotting stage. Instead of computing the theory there again. Do you agree? I could try to implement a version of that.

joezuntz (approved):

PR #164, opened 2021-02-09
Restore mean z in rail estimate

joezuntz:
This adds a post-RAIL calculation of the mean redshift of each galaxy.


empEvil (approved):
Looks fine to me, but maybe we need someone from RAIL to look at it too?

Issue #165, opened 2021-02-15
Use weighted means when calibrating


joezuntz:
We should use weighted means when computing selection and estimator calibrations in calibration_tools.py

joezuntz:
Addressed in #166

PR #166, opened 2021-02-16
Add weights to metacal selection bias calculation and tidy R_gamma part

joezuntz:
Addresses issue #165, adding weights to the selection bias calibration.
At the same time it simplifies the R_gamma (estimator) part of the calibration by using the ParallelMean tool.

The old version was also slightly wrong, since it did the weighted mean of the differences, instead of the diff of the weighted mean, which is not (quite) the same.

### DIFF
--- a/txpipe/test/test_cal.py
+++ b/txpipe/test/test_cal.py
@@ -11,7 +11,8 @@ def select_all_index(data):
     return np.arange(data['mcal_g2'].size)
 
 def select_all_where(data):
-    return np.where(data['mcal_g2'] != 68754337.86543434)
+    # we just want to select everything here too
+    return np.where(data['mcal_g2'] * 0 == 0)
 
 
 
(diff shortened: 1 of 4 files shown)


txpipe/utils/calibration_tools.py
```diff
@@ -164,10 +164,10 @@ def __init__(self, selector, delta_gamma):
             The difference in applied g between 1p and 1m metacal variants
```
joezuntz, line 164: It's probably easier to look at this class (`ParallelCalibratorMetacal`) in its entirety rather than the diff.

txpipe/test/test_cal.py
```diff
@@ -1,5 +1,94 @@
...
+
+
+def select_all_bool(data):
+    return np.repeat(True, data['mcal_g2'].size)
+
+def select_all_index(data):
+    return np.arange(data['mcal_g2'].size)
+
+def select_all_where(data):
+    return np.where(data['mcal_g2'] != 68754337.86543434)
```
jpratmarti, line 14: What is this number?

jpratmarti (approved):
I just left a comment/question. Otherwise looks ready to merge to me!

txpipe/test/test_cal.py
joezuntz: Just a random number that will never appear!  I'll rewrite to make more sensible.

jpratmarti (approved):

PR #169, opened 2021-02-23
Cosmodc2 magcut

jpratmarti:
Adapt `TXCosmoDC2Mock` stage to cut sources in magnitude, instead of snr.

### DIFF
--- a/txpipe/input_cats.py
+++ b/txpipe/input_cats.py
@@ -3,6 +3,7 @@ from .data_types import ShearCatalog, HDFFile
 from .utils.calibration_tools import band_variants, metacal_variants
 import numpy as np
 from .utils.timer import Timer
+import pdb
 
 class TXCosmoDC2Mock(PipelineStage):
     """
@@ -28,7 +29,8 @@ class TXCosmoDC2Mock(PipelineStage):
     config_options = {
         'cat_name':'cosmoDC2',
         'visits_per_band':165,  # used in the noise simulation
-        'snr_limit':4.0,  # used to decide what objects to cut out
+        'snr_limit': 4.0,  # used to decide what objects to cut out
+        'Mag_r_limit': -19, # used to decide what objects to cut out  
         'max_size': 99999999999999,  #for testing on smaller catalogs
         'extra_cols': "", # string-separated list of columns to include
         'max_npix':99999999999999,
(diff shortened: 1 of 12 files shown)


txpipe/input_cats.py
```diff
@@ -157,16 +159,20 @@ def run(self):
...
-            # Simulate the various output data sets
-            mock_photometry = self.make_mock_photometry(data)
 
             # Cut out any objects too faint to be detected and measured.
             # We have to do this after the photometry, so that we know if
             # the object is detected, but we can do it before making the mock
             # metacal info, saving us some time simulating un-needed objects
-            self.remove_undetected(data, mock_photometry)
+            
+            #self.remove_undetected(data, mock_photometry)
```
jpratmarti, line 168: This function is not necessary anymore if we do not cut in SNR. Also it was giving me a an error, that's why I commented. I also had to move `self.make_mock_photometry(data)` a few lines below to avoid errors.

txpipe/input_cats.py
```diff
@@ -28,12 +29,14 @@ class TXCosmoDC2Mock(PipelineStage):
     config_options = {
         'cat_name':'cosmoDC2',
         'visits_per_band':165,  # used in the noise simulation
-        'snr_limit':4.0,  # used to decide what objects to cut out
+        'snr_limit': 4.0,  # used to decide what objects to cut out
+        'Mag_r_limit': -19, # used to decide what objects to cut out
```
joezuntz, line 33: could you move this down so it's next to apply_mag_cut so that when I see it I'll realise it's not always applied?

txpipe/twopoint.py
```diff
@@ -272,11 +271,6 @@ def write_output(self, data, meta, results):
             theta = np.exp(d.object.meanlogr)
             npair = d.object.npairs
             weight = d.object.weight
-
```
joezuntz, line 275: Was this bit always wrong here?

txpipe/twopoint.py
```diff
@@ -532,20 +531,17 @@ def calculate_shear_shear(self, data, meta, i, j):
...
+            gg.process(cat_i, cat_j)
 
-
-        print(f"Rank {self.rank} calculating shear-shear bin pair ({i},{j}): {n_i} x {n_j} objects")
-
-        gg = treecorr.GGCorrelation(self.config)
-        gg.process(cat_i, cat_j)
-
+        print(f"Rank {self.rank} calculated shear-shear bin pair ({i},{j}): {n_i} x {n_j} objects")
+        print(gg.npairs, gg.weight, gg.xip, gg.xim)
```
joezuntz, line 544: can you remove this old print statement

txpipe/twopoint.py
```diff
@@ -581,30 +577,31 @@ def calculate_pos_pos(self, data, i, j):
...
+        nr = treecorr.NNCorrelation(self.config)
+        rr = treecorr.NNCorrelation(self.config)
+        
         if i==j:
-            cat_j = cat_i
-            rancat_j = rancat_i
             n_j = n_i
             n_rand_j = n_rand_i
+            nn.process(cat_i)
+            nr.process(cat_i, rancat_i)
```
joezuntz, line 589: In this case isn't nr the same as rn, so isn't this wasting time?

joezuntz (changes requested):
Just a few very minor changes - thanks!

txpipe/twopoint.py
jpratmarti: I would assume so but it was already written like this before by somebody else, not sure if they had their reasons, maybe @empEvil knows?

joezuntz: We can just do it for now, it's not a major thing, and we will need some changes for MPI treecorr anyway.

joezuntz (approved):

joezuntz:
Will merge once the tests are done.  Thanks so much!

txpipe/twopoint.py
jpratmarti: Actually I think you're right, before it was written like this since there was not a specific option for when the catalogs were equal. Let me quickly fix it since we're here.

PR #171, opened 2021-03-09
Miscellaneous utilities

joezuntz:
Adds two new misc utilities
- hex_escape, which replaces non-printable characters with a \x code so they can be printed to FITS files
- multi_where, which makes a dict of "where" results but potentially faster in some cases

It also makes use of the hex_escape function by applying it in the gitdiff function.


txpipe/test/test_utils.py
```diff
@@ -0,0 +1,34 @@
...
+def test_multi_where():
+    x = np.random.randint(0, 10, 100)
+    w = [1, 2, 3]
+
+    a = multi_where(x, w)
+
+    for m in w:
+        assert np.array_equal(np.where(x==m)[0], a[m])
+
+    assert a.keys() == set(w)
```
rmjarvis, line 34: What is the use case where this is faster than numpy?

I tested this unit test and timed `multi_where` vs `[np.where(x==m)[0] for m in w]` an found:
```
multi_where time =  5.602836608886719e-05
numpy where time =  2.1219253540039062e-05
```
So I thought, ok, it's just too small for the multi_where advantage to kick in.  Let's try bigger w and x:
```
x = np.random.randint(0, 100, 10000)
w = np.unique(np.random.randint(0,100,20))

t0 = time.time()
a = multi_where(x, w)
t1 = time.time()
b = {m: np.where(x==m)[0] for m in w}
t2 = time.time()

print('multi_where time = ',t1-t0)
print('numpy where time = ',t2-t1)
```
This yielded
```
multi_where time =  0.0028181076049804688
numpy where time =  0.0002880096435546875
```
So an even bigger advantage to numpy.  I'm sure you have something in mind, but I would guess that the (C-implemented) numpy function will usually win out over python loops.

txpipe/utils/__init__.py
```diff
@@ -1,4 +1,4 @@
 from .pixel_schemes import choose_pixelization, HealpixScheme, GnomonicPixelScheme
 from .number_density_stats import SourceNumberDensityStats, LensNumberDensityStats
-from .misc import array_hash, unique_list
+from .misc import array_hash, unique_list, multi_where
```
rmjarvis, line 3: And hex_escape, probably?

txpipe/test/test_utils.py
```diff
@@ -0,0 +1,34 @@
...
+    assert unique_list(x) == [1]
+
+    x = [1, 1, 2, 1, 1]
+    assert unique_list(x) == [1, 2]
+
+    x = [1, 1, 2, 1, 1, 2, 3, 3]
+    assert unique_list(x) == [1, 2, 3]
+
+    x = [-1, "cat", -1, "cat", "dog"]
+    assert unique_list(x) == [-1, "cat", "dog"]
```
rmjarvis, line 23: This is equivalent to `np.unique(x).tolist()`.  Not sure what the advantage of having an explicitly list-based version would be.  Most of the time `np.unique(x)` will do what you want, and if you absolutely need a list, adding `tolist()` is pretty straightforward.

rmjarvis (commented):
hex_escape seems good to me.

Not sure how useful unique_list and multi_where are.  It seems like the numpy equivalents are usually clear enough and typically faster.

joezuntz:
Hi Mike - thanks very much for the review!  Will put notes on the comments.

txpipe/utils/__init__.py
joezuntz: Added.

txpipe/test/test_utils.py
joezuntz: np.unique also sorts its outputs, which I didn't want for the use case of this - I wanted to maintain the order that some results are generated in but remove duplicates.  I couldn't see a numpy equivalent, without the sorting.

joezuntz: A quick check shows the time of the two to be pretty close, though both minuscule

```
from txpipe.utils import unique_list
import numpy as np
import time

# use case is tags for data points in sacc
# make array or random 4 letter words with a, b, c, d
letters = np.array(["a", "b", "c", "d"])
words = [np.random.choice(letters, size=4) for i in range(1000)]
words = np.array([''.join(w) for w in words])

# numpy version
t1 = time.time()
_, inds = np.unique(words, return_index=True)
inds.sort()
a = words[inds]

t2 = time.time()

b = unique_list(words)
t3 = time.time()
print('unique_list time = ',t3-t2)
print('numpy time = ',t2-t1)
```

txpipe/test/test_utils.py
joezuntz: The original use case was when the matches are heterogeneous, in particular because I was calling some tomographic bins 0, 1, 2, 3 and "all" (for the 2D case).  But you're right that it probably isn't worth the slow down, and in the end I was only using this in one place.  So I've removed it. Good point - thanks!

txpipe/test/test_utils.py
rmjarvis: OK.  I didn't look at the doc string of unique_list, but you should mention that it preserves order, unlike np.unique, if it doesn't already say so.

FWIW, here is a numpy issue requesting this functionality: https://github.com/numpy/numpy/issues/8621

joezuntz: Yep, it mentions it in the docstring.

PR #173, opened 2021-03-11
Splitters

joezuntz:
Adds a pair of tools, Splitter and DynamicSplitter, which will help out in a future PR changing data to make MPI low_mem treecorr work.

They take a stream of data and a bin index and then split into appropriate HDF5 file subgroups based on the bin values.

The fixed version works in parallel but requires you to know the sizes in advance.  The dynamic version is serial but does not have this limitation.

This PR follows on from the calibration utilities PR


txpipe/utils/splitters.py
```diff
@@ -0,0 +1,207 @@
...
+                dt = dtypes.get(col, "f8")
+                sub.create_dataset(col, (sz,), dtype=dt, maxshape=(None,))
+
+    def _size_check(self, b, e):
+        n = self.bin_sizes[b]
+
+        # Expand the columns by 50% if needed
+        while e > n:
+            sub = self.subgroups[b]
+            new_size = int(n * 1.5)
```
rmjarvis, line 192: I don't know how often e > n*1.5, but if that's ever a use case, you're probably better off either doing `new_size = int(e * 1.5)` to give space above the current new size requirement, or at least `new_size = max(e, int(n*1.5))` to avoid multiple resizes.

With either of these choices, you can switch the while to an if.

txpipe/test/test_splitters.py
```diff
@@ -0,0 +1,101 @@
...
+        splitter = DynamicSplitter(g, name, cols, counts, dtypes=dtypes)
+
+        for i in range(10):
+            s = i * 10
+            e = s + 10
+            for b in range(6):
+                w = np.where(bins[s:e]==b)
+                data = {"x": x[s:e][w], "y":y[s:e][w], "z":z[s:e][w]}
+                splitter.write_bin(data, b)
+        # does checks internally on size
```
rmjarvis, line 91: This one actually does something besides just checks.  It resizes the bins to the actual realized size.  Not sure if that just means you should edit this comment, or if there are interesting additional asserts to add related to that.  

E.g. Maybe could add asserts that the bin sizes are not right before the `finish` call, but are correct after.  To make sure the resizing step actually ran.

rmjarvis (approved):
Seems fine to me.  Just some minor comments.

The one thing that might be worth adding is something you have in your head, but I couldn't quite grok from this PR -- namely what is the actual use you have in mind for either of these classes.  I think it has to do with splitting up an input catalog into the patch catalogs that TreeCorr can then access from the multiple MPI processes.

TreeCorr natively would do that via
```
full_cat = treecorr.Catalog(full_file, **kwargs)
full_cat.write_patches('patch_dir')
del full_cat
mpi_cat = treecorr.Catalog(full_file, save_patch_dir='patch_dir', **kwargs)
# Each mpi job will load in the patches that it needs as it uses them.
```
So maybe if Splitter is designed to reproduce this (but presumably more efficiently by doing things in chunks), you could add a test that the Splitter output works correctly replacing the `write_patches` call.

txpipe/test/test_splitters.py
```diff
@@ -0,0 +1,101 @@
+from ..utils import Splitter, DynamicSplitter
+import numpy as np
+import tempfile
+import os
+
+
+# We can't test this under mockmpi because the
+# comm objects gets passed down to the C level
```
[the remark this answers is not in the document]
joezuntz: Thanks, yes, this is obsolete - originally I was going to have the parallelization via the comm dealt with inside the splitter, but I realised that it's not needed so moved it to the level above. Deleting.

txpipe/test/test_splitters.py
joezuntz: The sizes aren't necessarily wrong beforehand (depends on the RNG), but I'm adding a test that they are >= the final size.  The final size is checked implicitly by the np.allclose calls underneath.

txpipe/utils/splitters.py
joezuntz: Ha, you're absolutely right, that makes much more sense. Weird what blind spots I get sometimes.  I'm changing to return early if the size is already okay.

joezuntz:
Thanks Mike!  Re: the use case - this is mainly to split random and shear catalogs by tomographic bin, once they've been calibrated, so that they can be used directly from disc both by TreeCorr and in map-making. The actual use is in the next PR in the series (I split them up as they were otherwise too large to review properly).

PR #179, opened 2021-05-12
Aperture mass

alexander-mead:
Here is a first go. There will be many things that need changing, but I figured I'd at least start the pull request process.


joezuntz:
The diff for this will be much smaller once we merge the treecorr mpi branch in. This PR comes after that one.

joezuntz:
Ah, sorry, just noticed this is targeted at that branch already. Perfect!

examples/laptop_pipeline.yml
```diff
@@ -41,7 +41,8 @@ stages:
     - name: TXConvergenceMaps    # Make convergence kappa maps from g1, g2 maps
     - name: TXConvergenceMapPlots # Plot the convergence map
     - name: TXMapCorrelations    # plot the correlations between systematics and data
-
```
joezuntz, line 44: JZ: I undid some formatting changes to this file to avoid confusion

txpipe/twopoint.py
```diff
@@ -1,3 +1,4 @@
+from multiprocessing import Value
```
joezuntz, line 1: What does this do?

txpipe/twopoint.py
```diff
@@ -564,7 +565,113 @@ def read_metadata(self):
...
+    name='TXApertureMass'
+    inputs = [
+        ('calibrated_shear_catalog', ShearCatalog),
+        ('calibrated_lens_catalog', HDFFile),
+        ('shear_photoz_stack', HDFFile),
+        ('patch_centers', TextFile),
+        ('tracer_metadata', HDFFile),
+    ]
+    outputs = [
+        ('aperturemass_data', SACCFile),
```
joezuntz, line 580: Sorry to be finickity but could you put an underscore between aperture and mass, here and below?

txpipe/twopoint.py
```diff
@@ -564,7 +565,113 @@ def read_metadata(self):
...
+                    calcs.append((i,j,k))
+
+        if self.rank==0:
+            print(f"Running these calculations: {calcs}")
+
+        return calcs
+
+    def calculate_shear_shear(self, i, j):
+
+        gg = TXTwoPoint.calculate_shear_shear(self, i, j)
```
joezuntz, line 621: Can use `super()` here instead of `TXTwoPoint` in case we e.g. change the name of the parent class or otherwise modify the structure later.

txpipe/twopoint.py
```diff
@@ -564,7 +565,113 @@ def read_metadata(self):
...
+        gg.Map = gg.calculateMapSq() # MEAD: Rename gg.Map to something more informative
+
+        return gg
+
+    def write_output(self, source_list, lens_list, meta, results):
+
+        import sacc
+
+        # Names for aperture-mass correlation functions
+        MAPSQ = "mapsq"
```
joezuntz, line 631: I'd prefer to use longer more descriptive names for these (though this is a point of contention!).  Could you change to `galaxy_shear_apmass2` etc?

txpipe/twopoint.py
```diff
@@ -564,7 +565,113 @@ def read_metadata(self):
...
+
+        if self.rank==0:
+            print(f"Running these calculations: {calcs}")
+
+        return calcs
+
+    def calculate_shear_shear(self, i, j):
+
+        gg = TXTwoPoint.calculate_shear_shear(self, i, j)
+        gg.Map = gg.calculateMapSq() # MEAD: Rename gg.Map to something more informative
```
joezuntz, line 622: See your TODO here.

txpipe/twopoint.py
```diff
@@ -564,7 +565,113 @@ def read_metadata(self):
...
+        f.close()
+
+        # Now build up the collection of data points, adding them all to the sacc
+        for d in results:
+
+            # First the tracers and generic tags
+            tracer1 = f'source_{d.i}'
+            tracer2 = f'source_{d.j}'
+
+            theta = np.exp(d.object.meanlogr) # MEAD: Double check that this is correct for Map
```
joezuntz, line 657: I guess this is right by maybe worth checking as you say

joezuntz (changes requested):
A handful of cosmetic changes, otherwise looks good, thanks!

txpipe/twopoint.py
```diff
@@ -564,7 +565,113 @@ def read_metadata(self):
...
         return meta
 
-    
+# Aperture Mass class that inherits from TXTwoPoint
+class TXApertureMass(TXTwoPoint):
+
+    name='TXApertureMass'
+    inputs = [
+        ('calibrated_shear_catalog', ShearCatalog),
+        ('calibrated_lens_catalog', HDFFile),
```
joezuntz, line 574: I think the lens cat is no longer needed here.

txpipe/twopoint.py
alexander-mead: Hmm... I have no idea. I'm pretty sure I didn't add this. I'll delete it anyway since 'Value' doesn't seem to be used anywhere.

txpipe/twopoint.py
alexander-mead: Done

txpipe/twopoint.py
alexander-mead: Done

txpipe/twopoint.py
alexander-mead: I think that this is correct in as much as the same .meanlogr is 'correct' for standard galaxy shear. I suppose that generally one might want to integrate the theory over the (weighted) bin to compare with the measurement... probably overkill for most reasonable binning choices though.

txpipe/twopoint.py
alexander-mead: Done. Hopefully .mapsq is informative enough...

txpipe/twopoint.py
alexander-mead: If I comment this out then I get an error from `_read_nbin_from_tomography`, which seems to require 'calibrated_lens_catalog' to be present.

txpipe/twopoint.py
alexander-mead: Okay, I think I've implemented `super()` correctly now.

txpipe/twopoint.py
joezuntz: Ah, sorry, you're right..

alexander-mead:
I think I've addressed most/all of your comments now. I have left 4 comments in the code that start `MEAD:` with a few remaining details that might need to be addressed.

joezuntz (approved):

PR #184, opened 2021-06-28
Skysim 5000 analysis

joezuntz:
Tweaks to allow skysim 5000 to run faster

- adds saving into patches in TXTwoPoint
- adds a memory reporting function in the base class that was useful during dev
- fix for maps that go all the way to the south pole
- some fixes to Fourier version


joezuntz:
@chihway this was the issue you kindly volunteered to take over from @empEvil who is celebrating thesis submission and preparing emotionally for the upcoming Danish defeat by England in the European Championship.

jpratmarti (approved):

PR #190, opened 2021-07-16
Split lens and source photo-z catalogs

joezuntz:
This removes the assumption in the PZ parts of the pipeline that the source and lens catalogs come from the same underlying catalog.

This is in preparation for Metadetect where this will no longer be true.  After this I have to do something similar for the mapping stages and TXDiagnostics.

To cope with non-ugrizy data in RAIL this also updates to the newest version of that submodule.


txpipe/rail_pz.py
```diff
@@ -57,20 +65,66 @@ def run(self):
...
+    model_tag = "photoz_lens_model"
+
+    inputs = [
+        (training_tag, DataFile),
+        (testing_tag, DataFile),
+    ]
+
+    outputs = [(model_tag, PickleFile)]
+
+class PZRailTrainLensFromSource(PipelineStage):
```
chihway, line 110: this feels a little odd, maybe it's the name... like why is it not trainsourcefromlens? also, what if we only run part of 3x2, ie clustering only? (maybe that's a larger question)

txpipe/rail_pz.py
```diff
@@ -197,3 +270,74 @@ def write_output_chunk(self, output_file, start, end, z, pz_data):
         output_file["pdf/pdf"][start:end] = p
         output_file["point_estimates/z_mode"][start:end] = pz_data["zmode"]
         output_file["point_estimates/z_mean"][start:end] = mu
+
+
+
+class PZRailEstimateSource(PZRailEstimateLens):
```
chihway, line 276: is there a logic whether the source or the lens is the parent class in estimate vs. train?

txpipe/rail_pz.py
```diff
@@ -197,3 +270,74 @@ def write_output_chunk(self, output_file, start, end, z, pz_data):
...
+
+        # Also rename all the columns as they are loaded
+        return rename_iterated(it, renames)
+
+    def get_catalog_size(self):
+        with self.open_input("shear_catalog") as f:
+            nobj = f["shear/ra"].size
+        return nobj
+
+class PZRailEstimateSourceFromLens(PipelineStage):
```
chihway, line 329: same question as before

chihway:
looks good, just minor questions about naming, feel free to ignore if it makes no sense

examples/laptop_pipeline.yml
```diff
@@ -6,10 +6,14 @@ stages:
     - name: TXLensCatalogSplitter
     - name: TXStarCatalogSplitter
     - name: TXMeanLensSelector  # select objects for lens bins
-    - name: PZRailTrain
+    - name: PZRailTrainSource
       threads_per_process: 2
-    - name: PZRailEstimate
-    - name: TXPhotozStack        # stack p(z) into n(z)
+    - name: PZRailTrainLens
```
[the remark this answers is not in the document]
joezuntz: Done!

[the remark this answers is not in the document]
joezuntz: I've also modified to speed things up by using the same estimator for the two.

txpipe/rail_pz.py
joezuntz: Yeah, good idea, I've added the reverse stage SourceFromLens

txpipe/rail_pz.py
joezuntz: There wasn't, so I've harmonized this a bit.

joezuntz:
Thanks!  I've addressed your issues and also split rail off into two files in a new dir, but with no other changes.
I'll merge if the tests run okay.

PR #192, opened 2021-07-29
Stop assuming photometry and shear catalogs are the same size in map code

joezuntz:
The previous implementation of the mapping code assumed in some places that the photometry and shear catalogs were the same length; this won't be the case for metadetect and other catalogs.

The main map-making code already had optional separate lens and source mappers.

This PR:
- errors with a useful message if the user tries to use a combined mapper when the lengths are different
- split the auxiliary maps into aux_source_maps (PSF and flag maps) and aux_lens_maps (bright object and depth maps), with a stage to make each
- updates the downstream uses of the aux maps to use the right one(s)
- updates yaml pipeline and config files

Most of the changes are to configuration files and pipeline yaml, switching them to use the new stages; if you're reviewing this they don't really need to be gone through carefully!

### DIFF
--- a/txpipe/auxiliary_maps.py
+++ b/txpipe/auxiliary_maps.py
@@ -7,8 +7,8 @@ from .utils import choose_pixelization
 from .utils.calibration_tools import read_shear_catalog_type
 
 
-class TXAuxiliaryMaps(TXBaseMaps):
-    name = "TXAuxiliaryMaps"
+class TXAuxiliarySourceMaps(TXBaseMaps):
+    name = "TXAuxiliarySourceMaps"
     """
     This class generates:
         - depth maps
@@ -17,13 +17,12 @@ class TXAuxiliaryMaps(TXBaseMaps):
         - flag maps
     """
     inputs = [
-        ("photometry_catalog", HDFFile),  # for mags etc
         ("shear_catalog", ShearCatalog),  # for psfs
         ("shear_tomography_catalog", HDFFile),  # for per-bin psf maps
         ("source_maps", MapsFile),  # we copy the pixel scheme from here
     ]
     outputs = [
-        ("aux_maps", MapsFile),
+        ("aux_source_maps", MapsFile),
     ]
 
     config_options = {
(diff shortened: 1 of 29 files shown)


txpipe/maps.py
```diff
@@ -461,6 +464,17 @@ def data_iterator(self):
         # the source and lens map columns
         print("TODO: no lens weights here")
 
+        with self.open_input("photometry_catalog") as f:
```
arunkannawadi, line 467: The docstring for TXMainMaps says this is faster than running TXSourceMaps and then TXLensMaps. Is this still true with this check in place?

txpipe/auxiliary_maps.py
```diff
@@ -175,57 +130,129 @@ def finalize_mappers(self, pixel_scheme, mappers):
...
-            maps["aux_maps", f"flags/flag_{f}"] = (p, m)
+            maps["aux_source_maps", f"flags/flag_{f}"] = (p, m)
             # also print out some stats
             t = m.sum()
             print(f"Map shows total {t} objects with flag {f}")
 
         return maps
 
 
-class TXStandaloneAuxiliaryMaps(TXAuxiliaryMaps):
```
arunkannawadi, line 203: There is no `TXStandaloneAuxillaryMaps` any more. Is `TXAuxillaryLensMaps` supposed to substitute this altogether? It doesn't quite look that way to me.

arunkannawadi (approved):
Looks good to me, Joe. I haven't looked at any of the config changes, just the Python files. I have one comment and one question, aside from the inline ones:

1) Logically, it looks to me like TXAuxillarySourceMaps and TXAuxillaryLensMaps should inherit from TXSourceMaps and TXLensMaps instead of inherting from TXBaseMaps, but looking the code, I don't see an easy way. I don't have any useful suggestion here, but is it doable?

2) Is there a test code that I can run to see that TXPipe indeed handles source and lens catalogues with different sizes, or is that waiting on other PRs to be merged?

txpipe/maps.py
joezuntz: It should still be true yes, but the difference is minor and I've adjusted the docstring accordingly.

txpipe/auxiliary_maps.py
joezuntz: `TXStandaloneAuxillaryMaps` was a stub that was never actually filled in, so it never did anything in the end. TXAuxillaryLensMaps is the replacement for it, yes.

joezuntz:
Thanks so much for reviewing!

On your questions:
- There's some overlap here in the columns that the aux mapper and regular mapper load, but they are not exactly the same so I don't think there's a clear way to do it.  The behaviour common to both of them is inherited from TXBaseMaps
- Like you say, this is only one of several PRs that (should!) make the whole pipeline able to deal with completely separate source and lens cats from scratch. Once the next PR on diagnostic plots is in place I should be able to test this. I will add such a test to the CI for that PR.

PR #194, opened 2021-08-06
Stop assuming lens and source catalogs are the same size in diagnostics.

joezuntz:
This uses dask for the new lensing diagnostics and so requires ceci >= 1.5

### DIFF
--- a/txpipe/diagnostics.py
+++ b/txpipe/diagnostics.py
@@ -6,16 +6,14 @@ from .utils.fitting import fit_straight_line
 from .plotting import manual_step_histogram
 import numpy as np
 
-class TXDiagnosticPlots(PipelineStage):
+class TXSourceDiagnosticPlots(PipelineStage):
     """
     """
-    name='TXDiagnosticPlots'
+    name='TXSourceDiagnosticPlots'
 
     inputs = [
-        ('photometry_catalog', HDFFile),
         ('shear_catalog', ShearCatalog),
         ('shear_tomography_catalog', TomographyCatalog),
-        ('lens_tomography_catalog', TomographyCatalog),
     ]
 
     outputs = [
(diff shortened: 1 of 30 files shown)


txpipe/diagnostics.py
```diff
@@ -678,6 +672,146 @@ def plot_mag_histograms(self):
             fig.close()
 
 
+class TXLensDiagnosticPlots(PipelineStage):
```
joezuntz, line 675: Start of dask-related work

txpipe/diagnostics.py
```diff
@@ -678,6 +672,146 @@ def plot_mag_histograms(self):
             fig.close()
```
joezuntz, line 674: Start of dask-related work

txpipe/diagnostics.py
```diff
@@ -678,6 +672,146 @@ def plot_mag_histograms(self):
...
+        self.plot_snr_histograms(data, nbin)
+        self.plot_mag_histograms(data, nbin)
+
+        # These need to stay open until the end
+        for f in files:
+            f.close()
+
+
+    def load_data(self):
+        import dask.array as da
```
empEvil, line 718: Is there going to be an issue that this function returns dask arrays to a run that doesn't have dask loaded

txpipe/diagnostics.py
```diff
@@ -678,6 +672,146 @@ def plot_mag_histograms(self):
...
+
+    def plot_snr_histograms(self, data, nbin):
+        smin = self.config['snr_min']
+        smax = self.config['snr_max']
+        bins = np.geomspace(smin, smax, 50)
+        xlog = True
+        self.plot_histograms(data, nbin, 'snr', xlog, bins)
+
+    def plot_mag_histograms(self, data, nbin):
+        import dask.array as da
```
empEvil, line 758: why is it imported here, but not in the snr histograms?

empEvil (approved):
Have a few minor comments, but otherwise it looks fine

txpipe/diagnostics.py
joezuntz: Thanks - removed

txpipe/diagnostics.py
joezuntz: It should be okay - serial-mode dask is always loaded - in that context it basically works as just numpy. The parallel dask isn't always loaded. In general it's fine to pass things through other functions that don't have a module loaded locally.

PR #195, opened 2021-08-10
Switch CI to use RAIL BPZ algorithm

joezuntz:
This will switch the RAIL algorithm used in the continuous integration for the faster BPZ one.

### DIFF
--- a/examples/config/bpz_riz.columns
+++ b/examples/config/bpz_riz.columns
@@ -1,0 +1,11 @@
+#This file is used by DESC_BPZ to figure out which data columns correspond to which filter
+#and whether to include additional error convoluation or offsets.  See BPZ documentation
+#for further details
+#Filter        columns  cal(AB or Vega,optional)  zp_error  zp_offset
+DC2LSST_r      mag_r_lsst, mag_err_g_lsst   AB            0.005     0.0
+DC2LSST_i      mag_i_lsst, mag_err_i_lsst   AB            0.005     0.0
+DC2LSST_z      mag_z_lsst, mag_err_z_lsst   AB            0.005     0.0
+Z_S            redshift
+ID      id
+M_0     mag_i_lsst
+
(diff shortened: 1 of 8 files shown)


examples/config/bpz_riz.columns
```diff
@@ -0,0 +1,11 @@
+#This file is used by DESC_BPZ to figure out which data columns correspond to which filter
```
joezuntz, line 1: TODO: Generate these automatically.

examples/laptop_pipeline.yml
```diff
@@ -104,8 +104,6 @@ inputs:
     photometry_catalog: data/example/inputs/photometry_catalog.hdf5
     photoz_source_training: submodules/RAIL/tests/data/test_dc2_training_9816.hdf5
     photoz_source_testing: submodules/RAIL/tests/data/test_dc2_validation_9816.hdf5
-    photoz_lens_training: submodules/RAIL/tests/data/test_dc2_training_9816.hdf5
```
joezuntz, line 107: Removing this since the source and lens catalogs are the same in this example so we don't need to train both.

sschmidt23 (approved):
This looks good to me, though do you want zmax of 2.0 in the config.yml?  DC2 has galaxies up to z=3, so you are forcing those to a maximum photo-z of 2.0 which could cause a slight bias for higher z galaxies (with the benefit of decreased computational time of not checking the 2<z<3 redshift grid points)

I will comment that `zp_error` is added in quadrature to each band's error estimate.  This is done usually to account for "template mismatch error", i.e. the fact that discrete template sets are approximation of the underlying smooth galaxy distribution, and thus giving some extra "slosh" to the fluxes/colors will soften up the chi^2/likelihood a bit when galaxies fall in between the discrete templates.  It's definitely a little kludgey, so what numbers to use (and if they should be non-zero at all for a well calibrated template set and photometry) is pretty ad hoc.

But, since this is for testing, it should be fine, just bringing it up for future awareness

joezuntz:
Thanks so much!  I'll change the zmax to 3.  When I switch to generating the columns files I'll also allow the zero-points offsets and errors as config options, but that can come later.

PR #198, opened 2021-08-18
Fixing bug in the real space covariance 

jpratmarti:
In the old code the xip covariance was being computed twice, instead of computing once the xip and then the xim covariance. This pull request aims to fixing this.

### DIFF
--- a/txpipe/covariance.py
+++ b/txpipe/covariance.py
@@ -278,14 +278,15 @@ class TXFourierGaussianCovariance(PipelineStage):
             if xi_plus_minus1 != xi_plus_minus2:
                 # in the cross term, this contribution is subtracted.
                 # eq. 29-31 of https://arxiv.org/pdf/0708.0387.pdf
-                Bmode_F=-1 
+                Bmode_F=-1
+                print ("Subtracting")
             # below the we multiply zero to maintain the shape of the Cl array, these are effectively 
             # B-modes
             cov[1324] += np.outer(cl[13]*0 + SN[13], cl[24]*0 + SN[24]) * coupling_mat[1324] * Bmode_F
             cov[1423] += np.outer(cl[14]*0 + SN[14], cl[23]*0 + SN[23]) * coupling_mat[1423] * Bmode_F
 
         cov['final']=cov[1423]+cov[1324]
-
+        
         if self.do_xi:
             s1_s2_1 = self.get_spins(tracer_comb1)
             s1_s2_2 = self.get_spins(tracer_comb2)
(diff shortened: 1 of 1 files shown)


jpratmarti:
However, as the code stands now, the xip part stays the same as before, but the xim part gives NaN output. This should be fixed.

jpratmarti:
> However, as the code stands now, the xip part stays the same as before, but the xim part gives NaN output. This should be fixed.

This has been fixed now.

txpipe/covariance.py
```diff
@@ -402,21 +402,44 @@ def compute_covariance(self, cosmo, meta, two_point_data):
         cl_cache = {}
         xi_pm = [[('plus','plus'), ('plus', 'minus')], [('minus','plus'), ('minus', 'minus')]]
 
+        print("N2pt", N2pt)
```
joezuntz, line 405: Can you make the printed message more informative?

txpipe/covariance.py
```diff
@@ -402,21 +402,44 @@ def compute_covariance(self, cosmo, meta, two_point_data):
...
+
+
         for i in range(N2pt):
             tracer_comb1 = tracer_combs[i]
 
-            if i == N2pt0:
+            # reset variable in the new loop
+            count_xi_pm2 = 0 
+
+            if i == N2pt0-(N2pt-N2pt0):
```
joezuntz, line 422: This is a little confusing.  Can you phrase this as something like 
```
count_xi_pm1 = 1 if i in range(xim_start, xi_end) else 0
```
or similar, and the same for `count_xi_pm2` just so that it's more explicit?

txpipe/covariance.py
```diff
@@ -402,21 +402,44 @@ def compute_covariance(self, cosmo, meta, two_point_data):
...
+                if j >= N2pt0-(N2pt-N2pt0) and j<N2pt0:
+                    # this is the xim range
                     count_xi_pm2 = 1
 
-                if self.do_xi and ('source' in tracer_comb1) and ('source' in tracer_comb2):
+                if j >= N2pt0:
+                    # let's go back to xip
+                    count_xi_pm2 = 0
+
+                #print('i, j, tracer_comb1, tracer_comb2, xi_pm[count_xi_pm1][count_xi_pm2]:', i, j, tracer_comb1, tracer_comb2, xi_pm[count_xi_pm1][count_xi_pm2])
```
joezuntz, line 441: remove old commented line

txpipe/covariance.py
```diff
@@ -456,6 +479,7 @@ def compute_covariance(self, cosmo, meta, two_point_data):
                 # and fill it in, and the transpose component
                 cov_full[start_i:end_i, start_j:end_j] = cov_ij
                 cov_full[start_j:end_j, start_i:end_i] = cov_ij.T
+                #print("start_i, end_i, start_j, end_j", start_i, end_i, start_j, end_j)
```
joezuntz, line 482: remove old commented line

joezuntz (changes requested):

txpipe/covariance.py
jpratmarti: yes, thanks for the suggestion. I was so happy that it worked that didn't worry too much about the form!

joezuntz (approved):

PR #199, opened 2021-08-23
Adding variable for galaxy bias in covariance

jpratmarti:
Before galaxy_bias = 1 was hard-coded. This pull request allows to input the galaxy bias values the user wants in the config file.

### DIFF
--- a/examples/config/cosmodc2_config.yml
+++ b/examples/config/cosmodc2_config.yml
@@ -95,6 +95,9 @@ TXJackknifeCenters:
 TXDiagnosticPlots:
     shear_prefix: mcal_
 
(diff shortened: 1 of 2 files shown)


joezuntz:
I'm not sure how ceci copes with things that have None as the default - does it break being able to set them on the command line with a flag?  You could check by running `python -m TXPipe TXFourierGaussianCovariance --help` and seeing if it makes sense.

jpratmarti:
> python -m TXPipe TXFourierGaussianCovariance --help
@joezuntz in which environment should I do that? In the one I usually run ceci it doesn't work

joezuntz (approved):

PR #206, opened 2021-09-27
Trying to track down illegal instruction in github actions

joezuntz:
This isn't a real PR, I'm just trying to track down the cause of a crash in the CI that has just started happening.


joezuntz:
No joy. I just keep re-running things and eventually it works.

PR #207, opened 2021-09-29
Metadetection

joezuntz:
This PR adds metadetect calibration to the pipeline.

We don't replace metacal completely, but add an additional calibration system for metadetect.  We do modify the ingestion so that we create a metadetect-like catalog from what is actually a metacal catalog. (Need to do this for the GCR ingestion next). 

The process has also suggested some simplifications we can make throughout the pipeline.


txpipe/input_cats.py
```diff
@@ -228,12 +228,12 @@ def setup_photometry_output(self, photo_file, target_size):
 
 
 
-    def setup_metacal_output(self, metacal_file, target_size):
+    def setup_metadetect_output(self, metacal_file, target_size):
```
chihway, line 231: mix use of metadetect and metacal? I'm guessing both are using the same file structure so it doesn't matter, but is there a way to make this less confusing?

txpipe/source_selector.py
```diff
@@ -1,13 +1,43 @@
 from .base_stage import PipelineStage
 from .data_types import ShearCatalog, YamlFile, PhotozPDFFile, TomographyCatalog, HDFFile, TextFile
-from .utils import SourceNumberDensityStats
+from .utils import SourceNumberDensityStats, rename_iterated
 from .utils.calibration_tools import read_shear_catalog_type, apply_metacal_response
-from .utils.calibration_tools import metacal_variants, band_variants, MetacalCalculator, LensfitCalculator, HSCCalculator
+from .utils.calibration_tools import metacal_variants, metadetect_variants, band_variants, MetacalCalculator, LensfitCalculator, HSCCalculator, MetaDetectCalculator
+from .utils.calibrators import MetaCalibrator, MetaDetectCalibrator, LensfitCalibrator, HSCCalibrator
```
chihway, line 6: this is kind of random, but I wonder if HSCCalibrator should be HSMCalibrator (named after the shear measurement method)

txpipe/source_selector.py
```diff
@@ -28,8 +58,8 @@ class TXSourceSelector(PipelineStage):
     TODO: add option to use lensfit catalogs, which
     would be much much simpler.
     """
+    name = "TXSourceSelector"
```
chihway, line 61: is this still a viable stage? do we actually need to name it?

txpipe/source_selector.py
```diff
@@ -242,27 +228,31 @@ def apply_classifier(self, classifier, features, shear_data):
         """Apply the classifier to the measured magnitudes
         """
         bands = self.config['bands']
-        shear_prefix = self.config['shear_prefix']
 
         if self.config['shear_catalog_type'] == 'metacal':
-            variants = ['', '_1p', '_2p', '_1m', '_2m']
+            prefixes = ['mcal_', 'mcal_', 'mcal_', 'mcal_', 'mcal_']
```
chihway, line 233: oh wait, you still have the metacal prefix... so in the input_cat.py file was the metacal section missing?

txpipe/source_selector.py
```diff
@@ -364,20 +320,27 @@ def calculate_tomography(self, pz_data, shear_data, calibrators):
 
         # and calibrate the 2D sample.
         # This calibrator refers to self.select_2d
-        calibrators[-1].add_data(data)
+        calculators[-1].add_data(data)
 
         return tomo_bin, R, counts
 
+    def compute_per_object_response(self, data):
+        # The default implementation has no per-object weight
```
chihway, line 328: Is R response or weight?

txpipe/source_selector.py
```diff
@@ -673,9 +469,252 @@ def select_2d(self, data, calling_from_select=False):
...
+        R[:,0,0] = (data['mcal_g1_1p'] - data['mcal_g1_1m']) / delta_gamma
+        R[:,0,1] = (data['mcal_g1_2p'] - data['mcal_g1_2m']) / delta_gamma
+        R[:,1,0] = (data['mcal_g2_1p'] - data['mcal_g2_1m']) / delta_gamma
+        R[:,1,1] = (data['mcal_g2_2p'] - data['mcal_g2_2m']) / delta_gamma
+        return R
+
+    def apply_simple_redshift_cut(self, data):
+        # If we have the truth pz then we just need to do the binning once,
+        # as in the parent class
+        if not self.config['input_pz']:
```
chihway, line 526: do you mean true_z here instead of input_pz?

txpipe/source_selector.py
```diff
@@ -673,9 +469,252 @@ def select_2d(self, data, calling_from_select=False):
...
+    def setup_response_calculators(self, nbin_source):
+        delta_gamma = self.config['delta_gamma']
+        calculators = [MetaDetectCalculator(self.select, delta_gamma) for i in range(nbin_source)]
+        calculators.append(MetaDetectCalculator(self.select_2d, delta_gamma))
+        return calculators
+
+    def apply_simple_redshift_cut(self, data):
+        # If we have the truth pz then we just need to do the binning once,
+        # as in the parent class
+        if not self.config['input_pz']:
```
chihway, line 591: do you mean true_z instead of input_pz?

txpipe/source_selector.py
```diff
@@ -673,9 +469,252 @@ def select_2d(self, data, calling_from_select=False):
...
+        return pz_data
+
+
+    def setup_output(self):
+        outfile = super().setup_output()
+        n = outfile['tomography/source_bin'].size
+        nbin_source = outfile['tomography/source_counts'].size
+        group = outfile.create_group('response')
+        group.create_dataset('R', (nbin_source,2,2), dtype='f')
+        group.create_dataset('R_2d', (2,2), dtype='f')
```
chihway, line 616: do we not need the other quantities in the metacal output?

txpipe/source_selector.py
```diff
@@ -673,9 +469,252 @@ def select_2d(self, data, calling_from_select=False):
...
+        group.create_dataset('C', (nbin_source,2), dtype='f')
+        group.create_dataset('K_2d', (1,), dtype='f')
+        group.create_dataset('C_2d', (2), dtype='f')
+        return outfile
+
+
+    def compute_output_stats(self, calculator, mean, variance):
+        K, C, N = calculator.collect(self.comm, allgather=True)
+        calibrator = LensfitCalibrator(K, C)
+        mean_e = C
```
chihway, line 664: I didn't check how mean_e and C are used later, but would something like mean_e = C.copy() be less dangerous?

txpipe/utils/number_density_stats.py
```diff
@@ -32,14 +38,15 @@ def add_data(self, shear_data, shear_bin):
 
     def collect(self):
         # Get the basic shear numbers - means, counts, variances
-        variances = np.zeros((self.nbin_source, 2))
-        means = np.zeros((self.nbin_source, 2))
+        nb = self.nbin_source
+        variances = np.zeros((nb + 1, 2))
```
chihway, line 42: why is it +1? in the loop below looks like you only fill nb elements?

chihway:
I see Joe is updating things since I'm so slow in reviewing xd. In any case I went through the code once and left some minor comments. Also chatted with Joe to understand the main structural changes. 

I'm happy to take another fresh look at certain pieces of the code if useful before merging, but also fine for it to merge. Might be more efficient debugging with a more realistic catalog.

txpipe/input_cats.py
joezuntz: for the record - we chatted about this and the input_cats code now only generates metadetect-type files

txpipe/source_selector.py
joezuntz: In general I would agree, but I think there's a bunch of ways you might calibrate HSM. Whereas it's mainly only the one relevant version of the lensfit calibration.

txpipe/source_selector.py
joezuntz: I've added a note on this now - there is a check later to make sure that if anyone is still using the old name then it will complain and tell them how to change it.

txpipe/source_selector.py
joezuntz: The input_cat class only generates metadetect now, but we still support older metacal catalogs.

txpipe/source_selector.py
joezuntz: The response, e.g. the R_gamma matrix for metacal.

joezuntz: Just saw that the comment is wrong, thanks!

txpipe/source_selector.py
joezuntz: It should be `not input_z` which here is the same as `true_z`. I've tidied up and explained.

txpipe/source_selector.py
joezuntz: Yep - these are done in the superclass call at the top. I'll adda comment.

txpipe/source_selector.py
joezuntz: good idea - done.

txpipe/utils/number_density_stats.py
joezuntz: yep, but then we fill in the final one just afterwards with the 2D bit.

PR #208, opened 2021-10-15
Demo reorg

joezuntz:
Reorganize the example scripts and split up CI tests to make results clearer


examples/cosmodc2/pipeline_redmagic_no_shape_noise_full.yml
```diff
@@ -73,7 +73,7 @@ stages:
     - name: TXTwoPointPlotsFourier
```
empEvil, line 74: TXTwoPointLensCat seems to not exists as a stage

examples/2.2i/pipeline_1tract.yml
```diff
@@ -49,7 +49,7 @@ stages:
     - name: TXPSFDiagnostics
```
empEvil, line 50: Still uses TXPhotozStack which doesn't exist any longer

empEvil (commented):
I think it is slightly confusing that the examples folder also has the nersc and notebooks folders in them. All the other folders are different iterations of pipelines, but those two folders are not!

examples/cosmodc2/pipeline_redmagic_no_shape_noise_full.yml
joezuntz: fixed

examples/2.2i/pipeline_1tract.yml
joezuntz: also fixed!

joezuntz:
I've removed the old NERSC setup command and moved the notebooks folder to the top level.

empEvil (approved):

PR #213, opened 2021-10-18
separate source and lens noise mappers

joezuntz:


joezuntz:
Merged demo-reorg into this branch so this should only be looked at after that is merged.

joezuntz:
Already incorporated via another PR

PR #219, opened 2021-11-16
First version of TXNoiseMapsJax

Lhior:
It's a little funky but this is a first attempt at (lazily) utilizing GPU acceleration by implementing some Google Jax things into TXPipe. mpi4jax was tested with 2 GPUs and seemed to work fine but needs more testing. This currently breaks with clustering_realizations > 1 but will look into fixing that later. I'm using numpy as a generator for the first PRNG key seed for Jax but I'm open to suggestions on what to do to get a "random" starting point.

In the GPU cori nodes, using 1 GPU with chunk_rows:2000000 is about as fast as using 20 threads with the TXNoiseMaps regular stage.


txpipe/noise_maps.py
```diff
@@ -541,3 +541,216 @@ def finalize_mappers(self, pixel_scheme, mappers):
...
+            G2 = G2.at[index[pixels[sb_mask], source_bin[sb_mask], :]].add(g2r[sb_mask])
+            GW = GW.at[index[pixels[sb_mask], source_bin[sb_mask]]].add(weights[sb_mask])
+            lb_mask = sb_mask & (lens_bin >= 0)
+            ngal_split = ngal_split.at[index[pixels[lb_mask], lens_bin[lb_mask], jnp.arange(clustering_realizations), split[lb_mask]]].add(1)
+            #!!!!! Currently breaks with clustering_realizations > 1
+        # Sum everything at root
+        if self.comm is not None:
+            import mpi4jax
+            from mpi4py import MPI
+            mpi4jax.reduce(G1, MPI.SUM, root=0)
```
joezuntz, line 661: I think root=0 is the default? Fine to leave it if it's for clarity though.

txpipe/noise_maps.py
```diff
@@ -541,3 +541,216 @@ def finalize_mappers(self, pixel_scheme, mappers):
...
+            g2r = jnp.transpose(-sin * g1 + cos * g2)
+
+            pix_mask = pixels >= 0
+            sb_mask = (source_bin >= 0) & pix_mask
+            G1 = G1.at[index[pixels[sb_mask], source_bin[sb_mask], :]].add(g1r[sb_mask])
+            G2 = G2.at[index[pixels[sb_mask], source_bin[sb_mask], :]].add(g2r[sb_mask])
+            GW = GW.at[index[pixels[sb_mask], source_bin[sb_mask]]].add(weights[sb_mask])
+            lb_mask = sb_mask & (lens_bin >= 0)
+            ngal_split = ngal_split.at[index[pixels[lb_mask], lens_bin[lb_mask], jnp.arange(clustering_realizations), split[lb_mask]]].add(1)
+            #!!!!! Currently breaks with clustering_realizations > 1
```
joezuntz, line 656: Can you put a `# TODO:` here

joezuntz (changes requested):

txpipe/noise_maps.py
```diff
@@ -541,3 +541,216 @@ def finalize_mappers(self, pixel_scheme, mappers):
...
+        key = random.PRNGKey(np.random.randint(2**32))
+
+        # Loop through the data
+        for (s, e, data) in it:
+            print(f"Rank {self.rank} processing rows {s} - {e}")
+            # Send data to GPU
+            data = device_put(data)
+            source_bin = data['source_bin']
+            lens_bin = data['lens_bin']
+            ra = data['ra']
```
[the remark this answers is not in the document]
Lhior: That is correct. I think I'll just manually select what things to send to GPU instead of sending the whole of data since ra and dec are not used after that step.

txpipe/noise_maps.py
```diff
@@ -541,3 +541,216 @@ def finalize_mappers(self, pixel_scheme, mappers):
...
+            dec = data['dec']
+            orig_pixels = device_put(pixel_scheme.ang2pix(ra, dec))
+            pixels = device_put(index_map[orig_pixels])
```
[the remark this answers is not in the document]
Lhior: That seems like a good idea, I'll send it to GPU after the loop that adds the counters to it, that loop can also potentially be done in the same way I'm doing the [G1, G2, GW] but I think it's fast enough in its current implementation

txpipe/noise_maps.py
```diff
@@ -541,3 +541,216 @@ def finalize_mappers(self, pixel_scheme, mappers):
...
+        if self.rank == 0:
+            print("Saving maps")
+            outfile = self.open_output('source_noise_maps', wrapper=True)
+
+            # The top section has the metadata in
+            group = outfile.file.create_group("maps")
+            group.attrs['nbin_source'] = nbin_source
+            group.attrs['lensing_realizations'] = lensing_realizations
+            # Get outputs from GPU
+            g1 = device_get(g1)
```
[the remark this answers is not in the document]
Lhior: I just realized this! Thanks for pointing it out!

txpipe/noise_maps.py
```diff
@@ -541,3 +541,216 @@ def finalize_mappers(self, pixel_scheme, mappers):
...
+            key, subkey = random.split(key)
+            phi = random.uniform(subkey, shape=(lensing_realizations, n), minval=0, maxval=2*jnp.pi)
+            cos = jnp.cos(phi)
+            sin = jnp.sin(phi)
+            g1r = jnp.transpose(cos * g1 + sin * g2)
+            g2r = jnp.transpose(-sin * g1 + cos * g2)
+
+            pix_mask = pixels >= 0
+            sb_mask = (source_bin >= 0) & pix_mask
+            G1 = G1.at[index[pixels[sb_mask], source_bin[sb_mask], :]].add(g1r[sb_mask])
```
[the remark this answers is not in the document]
Lhior: Funnily enough, I did know about the `index` deprecation (and that's why `at` is being used here, I think I just forgot to remove it when swapping from `index_add`.

txpipe/noise_maps.py
Lhior: I think that's true in mpi4py but not sure in mpi4jax [https://mpi4jax.readthedocs.io/en/latest/api.html?highlight=reduce#mpi4jax.reduce](url), at least they show default values for `comm` and `token` but not `root`

txpipe/noise_maps.py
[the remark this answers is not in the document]
Lhior: I'll give that a try and do some benchmarking on Cori to see if it helps.

joezuntz (approved):

joezuntz:
@Lhior Can you check the changes I made (mostly on comments) and check you happy for me to merge this now?

Lhior:
@joezuntz Just reviewed all of your changes and they look good! Happy for you to merge this now!

txpipe/noise_maps.py
```diff
@@ -541,3 +541,291 @@ def finalize_mappers(self, pixel_scheme, mappers):
...
+
+        # Now we have finished looping through the data, we sum everything over the
+        # different processes to the root process
+        if self.comm is not None:
+            import mpi4jax
+            from mpi4py import MPI
+            mpi4jax.reduce(G1, MPI.SUM, root=0)
+            mpi4jax.reduce(G2, MPI.SUM, root=0)
+            mpi4jax.reduce(GW, MPI.SUM, root=0)
+            mpi4jax.reduce(ngal_split, MPI.SUM, root=0)
```
PhilipVinc, line 730: Hi @joezuntz , chiming in very randomly (I'm the author of `mpi4jax` and was recently looking at who uses the library for pure curiosity)

I noticed those few lines which are probably not what you want.

`mpi4jax` follows jax principles of being purely functional and not modifying in-place inputs. So those few lines are scheduling the reduction operation and they are  discarding the result.

I think you'd want to do something more like
```python
G1, token = mpi4jax.reduce(G1, MPI.SUM, root=0)
G2, token = mpi4jax.reduce(G2, MPI.SUM, root=0, token=token)
GW, token = mpi4jax.reduce(GW, MPI.SUM, root=0, token=token)
ngal_split, token = mpi4jax.reduce(ngal_split, MPI.SUM, root=0, token=token)
```

(the tokens are a way to enforce those operations to execute in a well-defined order, because jax does not necessarily execute operations in the same order you write them. This sometimes leads to MPI deadlocks if you do not use tokens. However, It might be that in this very simple case that's not needed.)

joezuntz: @PhilipVinc thanks so much, this is v useful!  Very kind of you to take the time to let us know this!

@Lhior could you open a new pull request to make these changes?

PhilipVinc: no probs.
By the way, just a suggestion: the block of logic above (the body of the loop for `for (s, e, data) in it:`) looks like could be written in a single `jax.jit` kernel.

Every jax operation that is not fitted has a fairly high overhead, so unless you have large > 10^6 elements you most likely are paying a considerable price. When jitting you pay the overhead only once.

PR #221, opened 2021-12-10
Fix fourier

jpratmarti:
Add functionality to compute `cl_noise` in `twopoint_fourier` from analytic function, to compare with the one obtained from the maps as a validation. Most of the code is copied from HSC branch, but some of the code has been modified to fix two bugs.

### DIFF
--- a/txpipe/maps.py
+++ b/txpipe/maps.py
@@ -494,7 +494,7 @@ class TXMainMaps(TXSourceMaps, TXLensMaps):
         # Still one mapper, but now we read both source and
         # lens maps from it.
         mapper, cal = mappers
-        pix, ngal, weighted_ngal, g1, g2, var_g1, var_g2, weights_g = mapper.finalize(self.comm)
+        pix, ngal, weighted_ngal, g1, g2, var_g1, var_g2, weights_g, esq = mapper.finalize(self.comm)
         maps = {}
 
         if self.rank != 0:
--- a/txpipe/maps.py
+++ b/txpipe/maps.py
@@ -349,7 +349,7 @@ class TXLensMaps(TXBaseMaps):
         # Again just the one mapper
         mapper = mappers[0]
         # Ignored return values are empty dicts for shear
-        pix, ngal, weighted_ngal, _, _, _, _, _ = mapper.finalize(self.comm)
+        pix, ngal, weighted_ngal, _, _, _, _, _, _ = mapper.finalize(self.comm)
         maps = {}
 
         if self.rank != 0:
--- a/txpipe/maps.py
+++ b/txpipe/maps.py
@@ -565,7 +565,10 @@ class TXDensityMaps(PipelineStage):
             ng[ng == healpy.UNSEEN] = 0
             delta_map = np.zeros(mask_copy.shape, dtype=np.float64)
             delta_map[mask_copy>0] = ng[mask_copy>0]/mask_copy[mask_copy>0]
-            delta_map[mask_copy>0] = delta_map[mask_copy>0]/np.mean(delta_map[mask_copy>0])-1
+            mu = np.mean(delta_map[mask_copy>0])
+            print(mu)
+            delta_map[mask_copy>0] = delta_map[mask_copy>0]/mu-1 
+                         
             density_maps.append(delta_map)
             '''
             mask_copy = mask.copy()
--- a/txpipe/maps.py
+++ b/txpipe/maps.py
@@ -279,7 +279,6 @@ class TXSourceMaps(TXBaseMaps):
             maps["source_maps", f"lensing_weight_{b}"] = (pix, weights_g[b])
             # added from HSC branch, to get analytic noise in twopoint_fourier
             out_e = np.zeros_like(esq[b])
-            #out_e[esq[b]>0] = esq[b][esq[b]>0]/weights_g[b][esq[b]>0]**2
             out_e[esq[b]>0] = esq[b][esq[b]>0]
             maps["source_maps", f"var_e_{b}"] = (pix, out_e)
 
@@ -561,7 +560,6 @@ class TXDensityMaps(PipelineStage):
             ng[np.isnan(ng)] = 0.0
             ng[ng == healpy.UNSEEN] = 0
             delta_map = np.zeros(mask.shape, dtype=np.float64)
-            #delta_map[mask_copy>0] = ng[mask_copy>0] # Assuming they are already weighted by the mask
             delta_map[mask>0] = ng[mask>0]/mask[mask>0] # Assuming that the weights do not include the mask
             mu = np.mean(delta_map[mask>0])
             delta_map[mask>0] = delta_map[mask>0]/mu-1
--- a/txpipe/maps.py
+++ b/txpipe/maps.py
@@ -565,7 +565,6 @@ class TXDensityMaps(PipelineStage):
             delta_map[mask>0] = ng[mask>0]/mask[mask>0] # Assuming that the weights do not include the mask
             mu = np.mean(delta_map[mask>0])
             delta_map[mask>0] = delta_map[mask>0]/mu-1
-            print(mu)
             density_maps.append(delta_map)
 
         # write output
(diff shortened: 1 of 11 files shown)


joezuntz:
Thanks for this! It looks like there are a few other places you need to update to add esq. You can find them by clicking "Details" next to the failing tests and then expanding the "Show logs" arrow to find what crashed and where.

joezuntz:
Looks like the tests pass - is this ready for PR now?

jpratmarti:
> Looks like the tests pass - is this ready for PR now?

great! Yes @joezuntz

jpratmarti:
@joezuntz Weirdly enough the ConvergenceMapsPlots fails in this tests too.. Do you have any ideas why or how to fix it?

txpipe/maps.py
```diff
@@ -553,20 +558,15 @@ def run(self):
...
-            d = (ng - mu) / mu
-            # remove nans
-            d[mask == 0] = 0
-            density_maps.append(d)
+            delta_map = np.zeros(mask.shape, dtype=np.float64)
+            #delta_map[mask_copy>0] = ng[mask_copy>0] # Assuming they are already weighted by the mask
+            delta_map[mask>0] = ng[mask>0]/mask[mask>0] # Assuming that the weights do not include the mask
+            mu = np.mean(delta_map[mask>0])
+            delta_map[mask>0] = delta_map[mask>0]/mu-1
+            print(mu)
```
fjaviersanchez, line 568: you probably want to remove the print statement or add some description about what is printed (sorry, this is probably my fault).

txpipe/masks.py
```diff
@@ -22,7 +21,7 @@ def run(self):
             bright_obj = f.read_map("bright_objects/count")
             depth = f.read_map("depth/depth")
             pixel_scheme = choose_pixelization(**metadata)
-
+
```
fjaviersanchez, line 24: The linter is going to complain about this line. Either remove or remove the blank spaces on the left.

fjaviersanchez (changes requested):
Changes look good to me. There a couple of minor things, but otherwise it's good. Thanks!

txpipe/masks.py
jpratmarti: Removed!

txpipe/maps.py
jpratmarti: Removed!

fjaviersanchez (approved):
Looks good to me, thanks @jpratmarti!

txpipe/maps.py
```diff
@@ -277,6 +277,11 @@ def finalize_mappers(self, pixel_scheme, mappers):
             maps["source_maps", f"var_g1_{b}"] = (pix, var_g1[b])
             maps["source_maps", f"var_g2_{b}"] = (pix, var_g2[b])
             maps["source_maps", f"lensing_weight_{b}"] = (pix, weights_g[b])
+            # added from HSC branch, to get analytic noise in twopoint_fourier
+            out_e = np.zeros_like(esq[b])
+            #out_e[esq[b]>0] = esq[b][esq[b]>0]/weights_g[b][esq[b]>0]**2
```
joezuntz, line 282: Can you remove the old commented-out code here?

txpipe/maps.py
```diff
@@ -553,20 +558,14 @@ def run(self):
...
-            # mean clustering galaxies per pixel in this map
-            mu = np.average(ng, weights=mask_copy)
-            print(f"Mean number density in bin {i} = {mu}")
-            # and then use that to convert to overdensity
-            d = (ng - mu) / mu
-            # remove nans
-            d[mask == 0] = 0
-            density_maps.append(d)
+            delta_map = np.zeros(mask.shape, dtype=np.float64)
+            #delta_map[mask_copy>0] = ng[mask_copy>0] # Assuming they are already weighted by the mask
```
joezuntz, line 564: also remove old code here

txpipe/metadata.py
```diff
@@ -138,7 +138,8 @@ def read_area(self):
             m = f.read_map("mask")
             pixel_scheme = choose_pixelization(**f.read_map_info("mask"))
 
-        num_hit = (m > 0).sum()
+        #num_hit = (m > 0).sum()
```
joezuntz, line 141: and here!

txpipe/noise_maps.py
```diff
@@ -184,17 +185,31 @@ def run(self):
                 for i in range(clustering_realizations):
                     # We have computed the first half already,
                     # and we have the total map from an earlier stage
-                    half1 = ngal_split[:, b, i, 0]
-                    half2 = ngal_split[:, b, i, 1]
-
+                    #half1 = ngal_split[:, b, i, 0]
```
joezuntz, line 188: more old code :-)

txpipe/noise_maps.py
```diff
@@ -184,17 +185,31 @@ def run(self):
...
+                        half1 =	(ngal_split[:, b, i, 0])/mask[pixels]
+                        half2 = (ngal_split[:, b, i, 1])/mask[pixels]
+                    
                     # Convert to overdensity.  I thought about
                     # using half the mean from the full map to reduce
                     # noise, but thought that might add covariance
                     # to the two maps, and this shouldn't be that noisy
-                    mu1 = np.average(half1, weights=mask[pixels])
-                    mu2 = np.average(half2, weights=mask[pixels])
+                    #mu1 = np.average(half1, weights=mask[pixels])
```
joezuntz, line 205: I'll stop going through this now finding all the old commented out bits, but in general can you remove them?

txpipe/noise_maps.py
```diff
@@ -184,17 +185,31 @@ def run(self):
...
+                    #mu2 = np.average(half2, weights=mask[pixels])
                     # This will produce some mangled sentinel values
                     # but they will be masked out
+
+                    mu1 = np.mean(half1)
```
joezuntz, line 210: Is there a reason we don't want the weights here?  I haven't thought this through though!

txpipe/noise_maps.py
```diff
@@ -184,17 +185,31 @@ def run(self):
...
-
+                    #half1 = ngal_split[:, b, i, 0]
+                    #half2 = ngal_split[:, b, i, 1]
+
+                    half1 = np.zeros(npix)
+                    half2 = np.zeros_like(half1)
+
+                    if self.config['mask_in_weights']:
+                        half1 = ngal_split[:, b, i, 0]
+                        half2 =	ngal_split[:, b, i, 1]
```
joezuntz, line 196: funny spacing

txpipe/twopoint_fourier.py
```diff
@@ -546,12 +547,33 @@ def compute_power_spectra(self, pixel_scheme, i, j, k, maps, workspaces, ell_bin
 
         workspace = workspaces[(i,j,k)]
 
-        # Get the coupled noise C_ell values to give to the master algorithm
-        cl_noise = self.compute_noise(i, j, k, ell_bins, maps, workspace)
-
-        # Run the master algorithm
-        c = nmt.compute_full_master(field_i, field_j, ell_bins,
-            cl_noise=cl_noise, cl_guess=cl_guess, workspace=workspace, n_iter=1)
+        if self.config['analytic_noise']==False:
```
joezuntz, line 550: I usually prefer to minimize the mental logical people have to do and just do `if self.config['analytic_noise']:` (which would mean swapping the order here).

joezuntz (changes requested):

txpipe/noise_maps.py
jpratmarti: @fjaviersanchez what was the reason for removing the weights here?

fjaviersanchez: Oh, sorry! It's because I am assuming that the only weights are due to the mask and that's taken into account in lines 198 and 199. If there are additional weights we would have to use `np.average` with the weights. For the HSC branch, so far, we only have the mask and no weights (we use deprojection instead) so the weights were not relevant. I think that for the general case where you have non-unity weights for all galaxies, you would need to bring that weighted average back.

jpratmarti: Thanks Javi! OK, I'll revert these changes then.

txpipe/twopoint_fourier.py
jpratmarti: Done

txpipe/noise_maps.py
jpratmarti: Done

PR #224, opened 2022-01-07
Gaussian sims

jpratmarti:
Changes to be able to run through some simpler simulation.

### DIFF
--- a/txpipe/theory.py
+++ b/txpipe/theory.py
@@ -73,7 +73,12 @@ class TXTwoPointTheoryReal(PipelineStage):
             nz = smooth_nz(Ti.nz) if smooth else Ti.nz
             print("smooth:", smooth)
             # Convert to CCL form
-            tracers[name] = pyccl.WeakLensingTracer(cosmo, (Ti.z, nz))
+            try:
+                tracers[name] = pyccl.WeakLensingTracer(cosmo, (Ti.z, nz))
+            except:
+                print('To avoid a CCL_ERROR_INTEG we reduce the number of points in the nz by half in source bin %d'%i)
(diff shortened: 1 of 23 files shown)


yomori (approved):

txpipe/theory.py
```diff
@@ -73,7 +73,12 @@ def get_ccl_tracers(self, s, cosmo, smooth=False):
             nz = smooth_nz(Ti.nz) if smooth else Ti.nz
             print("smooth:", smooth)
             # Convert to CCL form
-            tracers[name] = pyccl.WeakLensingTracer(cosmo, (Ti.z, nz))
+            try:
+                tracers[name] = pyccl.WeakLensingTracer(cosmo, (Ti.z, nz))
+            except:
```
joezuntz, line 78: again, good to just catch the one error

txpipe/utils/theory.py
```diff
@@ -92,7 +92,13 @@ def theory_3x2pt(cosmology_file, tracers, nbin_source, nbin_lens, fourier=True):
     for i in range(nbin_source):
         x = tracers[f'source_{i}']
         tag = ('S', i)
-        CTracers[tag] = ccl.WeakLensingTracer(cosmo, dndz=(x.z, x.nz))
+        try:
+            CTracers[tag] = ccl.WeakLensingTracer(cosmo, dndz=(x.z, x.nz))
+        except:
```
joezuntz, line 97: clearly we should centralize this better! But could change this bit too.

joezuntz (changes requested):

txpipe/theory.py
jpratmarti: How can I specify the error? I tried it but I was not sure how to do it. Is it using NameError?

joezuntz (approved):

PR #226, opened 2022-01-19
Update HSC branch to compute covariances with TJPCov

carlosggarcia:
I have merged the tjpcov-related files from master to the HSC branch.

### DIFF
--- a/txpipe/twopoint_fourier.py
+++ b/txpipe/twopoint_fourier.py
@@ -113,6 +113,7 @@ class TXTwoPointFourier(PipelineStage):
         # TODO: set ell binning from config
         ell_bins = self.choose_ell_bins(pixel_scheme, f_sky)
 
+        self.hash_metadata = None  # Filled in make_workspaces
         workspaces = self.make_workspaces(maps, calcs, ell_bins)
 
         # Load the n(z) values, which are both saved in the output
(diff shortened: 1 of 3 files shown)


txpipe/twopoint_fourier.py
```diff
@@ -222,44 +226,57 @@ def load_maps(self):
...
             systmaps_path = pathlib.Path(systmaps_clustering_dir)
             for systmap in systmaps_path.iterdir():
-                systmap_file = str(systmap)
-                print('Reading clustering systematics map file:', systmap_file)
                 try:
                     if systmap.is_file():
-                        syst_map = healpy.read_map(systmap_file, verbose=False)
-                    else:
-                        warnings.warn(f'Map from {systmap} is missing. Skipping...')
+                        if pathlib.Path(systmap).suffix != ".fits":
```
fjaviersanchez, line 242: I think that this would break if the map is compressed? Maybe add `or pathlib.Path(systmap).suffix != ".fits.gz"`

fjaviersanchez (approved):
Thanks @carlosggarcia! It looks good to me. There's a potential problem if we save maps in `fits.gz` format, but I think that's not the case so far, so feel free to just merge this, and we can open an issue later if this causes problems (also easier to track/fix just that problem).

txpipe/twopoint_fourier.py
damonge: There are probably big savings using `fits.gz` for HSC

carlosggarcia: @fjaviersanchez Done. In fact, I would default to `fits.gz` maps.

PR #228, opened 2022-01-24
Update HSC branch to use already computed noise in TJPCov

carlosggarcia:
Instead of recomputing the noise in TJPCov, we pass the noise computed already by TXPipe. 

I have also fixed:

- an error introduced in previous update: I had removed the possibility of using ell_edges for the binning. Now it can be used again as it was. Note that in the way that is (and was) implemented, your last ell_edge must be 3 * nside for the covariance to be computed. 
- the shear analytical noise: Nell = 0 for ell = 1, 2 now


fjaviersanchez (approved):
Changes look good. Thanks a lot @carlosggarcia

PR #229, opened 2022-01-26
Use computed coupled noise in TXTwoPointFourier to compute covariance with TJPCov

carlosggarcia:
I have done the following:
-  Store the noise and coupled noise in the `twopoint_fourier.fits` file. The noise as a tag in each of the data points (copied from the HSC branch) and the coupled noise as a metadata of the tracers. This assumes the coupled noise to be constant and might be necessary to be changed in the future.
- Fix the analytic noise computation. The expressions used give directly the coupled noise not the decoupled. I haven't checked the actual calculation, just assumed it was OK and interpret it as a coupled noise, instead of decouple. I think @damonge agrees with this. @jpratmarti it'd be great if you could have a look at this, since you've been working on it before.

### DIFF
--- a/txpipe/covariance.py
+++ b/txpipe/covariance.py
@@ -658,7 +658,9 @@ class TXFourierTJPCovariance(PipelineStage):
         calculator = tjpcov.main.CovarianceCalculator({"tjpcov":tjp_config})
 
         cache = {'workspaces': workspaces}
-        covmat = calculator.get_all_cov_nmt(cache=cache)
+        tracer_nlcp = self.get_tracer_noise_from_sacc(cl_sacc)
+        covmat = calculator.get_all_cov_nmt(cache=cache,
+                                            tracer_noise_coupled=tracer_nlcp)
 
         # Write the sacc file with the covariance
         if self.rank == 0:
@@ -666,6 +668,17 @@ class TXFourierTJPCovariance(PipelineStage):
             output_filename = self.get_output('summary_statistics_fourier')
             cl_sacc.save_fits(output_filename, overwrite=True)
             print("Saved power spectra with its Gaussian covariance")
+
+    def get_tracer_noise_from_sacc(self, cl_sacc):
+        # This could be done inside TJPCov:
+        # https://github.com/LSSTDESC/TJPCov/issues/31
+
+        tracer_noise = {}
+        for trn, tr in cl_sacc.tracers.items():
+            if 'nl_cp' in tr.metadata:
+                tracer_noise[trn] = tr.metadata['nl_cp']
+
+        return tracer_noise
 
     def get_workspaces_dict(self, cl_sacc, masks_names):
         # Based on txpipe/twopoint_fourier.py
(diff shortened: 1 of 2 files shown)


jpratmarti (approved):

joezuntz:
Thanks! I'd like to have a quick look at this tomorrow.

txpipe/covariance.py
```diff
@@ -658,7 +658,9 @@ def run(self):
         calculator = tjpcov.main.CovarianceCalculator({"tjpcov":tjp_config})
 
         cache = {'workspaces': workspaces}
-        covmat = calculator.get_all_cov_nmt(cache=cache)
+        tracer_nlcp = self.get_tracer_noise_from_sacc(cl_sacc)
+        covmat = calculator.get_all_cov_nmt(cache=cache,
```
joezuntz, line 662: Will this function cope if the tracer_nlcp dictionary is empty?

txpipe/covariance.py
```diff
@@ -667,6 +669,17 @@ def run(self):
             cl_sacc.save_fits(output_filename, overwrite=True)
             print("Saved power spectra with its Gaussian covariance")
 
+    def get_tracer_noise_from_sacc(self, cl_sacc):
+        # This could be done inside TJPCov:
+        # https://github.com/LSSTDESC/TJPCov/issues/31
+
+        tracer_noise = {}
+        for trn, tr in cl_sacc.tracers.items():
+            if 'nl_cp' in tr.metadata:
```
joezuntz, line 678: Could you give it a more descriptive name than nl_cp, so newcomers can tell what it means?

This mainly applies to the name it has in the data file, but also useful in some places in the code itself as well.  e.g. `n_ell` and `n_ell_coupled` maybe?

joezuntz (changes requested):

txpipe/covariance.py
carlosggarcia: `tracer_nlp` should never be empty; i.e. should be always taken from what `TXTwoPointFourier` computes. In any case, I think it will not fail (but I think the expressions in TJPCov for the noise are different from those in TXPipe) because when passing the noise it checks first the tracer is in the dictionary: https://github.com/LSSTDESC/TJPCov/blob/01f6620d28d4c788eefbb1effd62139a22dcf3c8/tjpcov/main.py#L676-L677

txpipe/covariance.py
carlosggarcia: Done

txpipe/covariance.py
carlosggarcia: I've added a warning in case one of the tracers metadata `n_ell_coupled` is missing. I'm tempted to raise an error. I think it would be the safest option.

joezuntz: Thanks! If there's never a reason that n_ell_coupled could be missing (e.g. are there any simple tests people might like to do where it's not there?) then an error makes sense, yep!

carlosggarcia: At the moment, the code always computes the noise. So, it will always have `n_ell_coupled`. If this is modified later, one can add a flag or something. I've changed to an error.

joezuntz (approved):

PR #232, opened 2022-02-11
Gaussian sims run2

jpratmarti:
Additional fixes and bugs we are finding when running through the gaussian sims and also CosmoDC2 with metadetect.


txpipe/theory.py
```diff
@@ -103,81 +110,85 @@ def replace_with_theory_real(self, s, cosmo):
         ell = np.unique(np.logspace(np.log10(2), 5, 400).astype(int))
         tracers = self.get_ccl_tracers(s, cosmo)
 
-        for i in range(nbin_source):
-            for j in range(i + 1):
-                print(f"Computing theory lensing-lensing ({i},{j})")
+        if "galaxy_shear_xi_plus" and "galaxy_shear_xi_minus" in s.get_data_types():
```
joezuntz, line 113: This line isn't doing what you think it's doing. `"galaxy_shear_xi_plus" and "galaxy_shear_xi_minus"` will always evaluate to `"galaxy_shear_xi_plus"`, for weird python reasons. You need `"galaxy_shear_xi_plus" in s.get_data_types() and "galaxy_shear_xi_minus" in s.get_data_types()` or equivalent.  Or you could just assume that if xip exists then so does xim.

txpipe/theory.py
```diff
@@ -103,81 +110,85 @@ def replace_with_theory_real(self, s, cosmo):
...
+
+                    # compute theory
+                    print(tracers[f"source_{i}"], tracers[f"source_{j}"])
+                    cl = pyccl.angular_cl(
+                        cosmo, tracers[f"source_{i}"], tracers[f"source_{j}"], ell
+                    )
+                    theta, *_ = s.get_theta_xi(
+                        "galaxy_shear_xi_plus", f"source_{i}", f"source_{j}"
+                    )
+                    xip = pyccl.correlation(cosmo, ell, cl, theta / 60, corr_type="L+")
```
joezuntz, line 126: are the units right here? I have no idea what pyccl wants or what we're saving things in!  Worth a comment to clarify.

txpipe/theory.py
```diff
@@ -103,81 +110,85 @@ def replace_with_theory_real(self, s, cosmo):
...
-            for j in range(i + 1):
-                print(f"Computing theory density-density ({i},{j})")
-
-                # compute theory
-                cl = pyccl.angular_cl(
-                    cosmo, tracers[f"lens_{i}"], tracers[f"lens_{j}"], ell
-                )
-                theta, *_ = s.get_theta_xi(
-                    "galaxy_density_xi", f"lens_{i}", f"lens_{j}"
+                    cosmo, tracers[f"lens_{i}"], tracers[f"lens_{i}"], ell
```
joezuntz, line 148: It looks like this first set of cl aren't actually used, since the ones in the loop are instead?

txpipe/twopoint_fourier.py
```diff
@@ -732,9 +731,20 @@ def compute_noise_analytic(self, i, j, k, maps, f_sky, workspace):
             ndens = (
                 metadata["tracers/lens_density"][i] * 3600 * 180 / np.pi * 180 / np.pi
             )
+            with self.open_input("mask", wrapper=True) as f:
+                info = f.read_map_info("mask")
```
joezuntz, line 735: Info variable is not needed here.

Also we could also pass the map in as a variable to avoid reading it again for every bin pair.

joezuntz (changes requested):

txpipe/theory.py
jpratmarti: oh I see, interesting! Fixed.

txpipe/theory.py
jpratmarti: CCL gets degrees in that function, so it should be correct. Added a comment.

txpipe/theory.py
jpratmarti: True! Good catch, I removed that.

txpipe/twopoint_fourier.py
jpratmarti: Yes, I removed info implemented your suggestion of passing it as a variable.

joezuntz (approved):

fjaviersanchez (approved):
Things look good to me! Thanks a lot for the work.

PR #238, opened 2022-03-10
Run gaussian sims3

jpratmarti:
More bug fixes, improvements and config setting changes to run through the Gaussian simulations with LSST Y1 like settings.


joezuntz:
Looks good! Ready to merge now?

jpratmarti:
> Looks good! Ready to merge now?

We can either merge now or else wait to see if there are more changes? I am not sure what is best practice to do smaller mergers and less branches or the other way around. There is a bug fix in the usual Fourier space covariance that Travis did not catch though that might be worth merging sooner rather than later.

PR #240, opened 2022-03-16
Add number of patches to file metadata

### LINKED ISSUE #176: TwoPoint metadata should include the number of patches it uses for jackknifing
Currently the number of jackknife patches used by TwoPoint is not saved in the metadata, it probably should be.

joezuntz:
Saves the number of patches as metadata in the sacc file output when present.

Closes #176

### DIFF
--- a/txpipe/twopoint.py
+++ b/txpipe/twopoint.py
@@ -427,6 +427,11 @@ class TXTwoPoint(PipelineStage):
             A list of (bin1, bin2, bin_type) where bin1 and bin2 are indices
             or bin labels and bin_type is one of the constants SHEAR_SHEAR,
             SHEAR_POS, or POS_POS.
+
+        meta: dict
+            A dict to which the number of patches (or zero, if no patches) will
+            be added for each catalog type, with keys "npatch_shear", "npatch_pos",
+            and "npatch_ran".
         """
         # Make the full list of catalogs to run
         cats = set()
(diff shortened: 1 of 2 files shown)


txpipe/twopoint.py
```diff
@@ -414,7 +414,7 @@ def call_treecorr(self, i, j, k):
         sys.stdout.flush()
         return result
 
-    def prepare_patches(self, calcs):
+    def prepare_patches(self, calcs, meta):
```
empEvil, line 417: The documentation has not been updated

empEvil (changes requested):
Looks good, its a minor issue that we should update the documentation for the changes!

txpipe/twopoint.py
joezuntz: Thanks, I've fixed!

txpipe/utils/patches.py
```diff
@@ -241,23 +241,24 @@ def run(cls, cat, chunk_rows, comm=None):
                 print(
                     f"Catalog {cat.file_name} does not have a patch directory set, so not making patches."
                 )
-            return
+            return 0
```
empEvil, line 244: It should probably also be mentioned in the documentation for this class that it now returns a scalar value!

empEvil (approved):
Great!

PR #242, opened 2022-03-21
Gaussian sims run4

jpratmarti:


txpipe/twopoint_fourier.py
```diff
@@ -551,7 +555,7 @@ def compute_power_spectra(
 
         # We need the theory spectrum for this pair
         # TODO: when we have templates to deproject, use this.
-        theory = cl_theory[(i, j, k)]
+        theory = cl_theory[(i, j, k)] # this is not needed anymore. Should I remove it?
```
carlosggarcia, line 558: What was that for?

txpipe/twopoint_fourier.py
```diff
@@ -844,23 +863,45 @@ def save_power_spectra(self, tracers, nbin_source, nbin_lens):
...
+                        # If computed through simulations, it might be better to
+                        # take the mean since, for now, only a float can be passed
+                        i = 0 if "lens" in tracer1 else 2
+                        tr.metadata["n_ell_coupled"] = np.mean(d.noise_coupled)
+                    else:
+                        # Save the last element because the first one is zero for
+                        # shear
+                        tr.metadata["n_ell_coupled"] = d.noise_coupled[-1]
+
+
```
carlosggarcia, line 902: To avoid having two ways of writing `tr.metadata["n_ell_coupled"]`, which potentially could lead to bugs, I would leave that part as it was before and modify the `n_ell_coupled` in this line. 

In other words, I would add in this line
```
if self.config["gaussian_sims_factor"] != [1.] and "lens" in tracer1:  
    print (tracer1)
    print("ATTENTION: We are multiplying the coupled noise saved in the sacc file by the gaussian sims factor.")
    print("Original noise:", d.noise_coupled)
    tr.metadata["n_ell_coupled"] *= self.config["gaussian_sims_factor"][int(tracer1[-1])]**2
```

txpipe/twopoint_fourier.py
```diff
@@ -844,23 +863,45 @@ def save_power_spectra(self, tracers, nbin_source, nbin_lens):
...
+                    if self.config["analytic_noise"] is False:
+                        # If computed through simulations, it might be better to
+                        # take the mean since, for now, only a float can be passed
+                        i = 0 if "lens" in tracer1 else 2
+                        tr.metadata["n_ell_coupled"] = np.mean(d.noise_coupled*self.config["gaussian_sims_factor"][int(tracer1[-1])]**2)
+                    else:
+                        # Save the last element because the first one is zero for
+                        # shear
+                        tr.metadata["n_ell_coupled"] = d.noise_coupled[-1]*self.config["gaussian_sims_factor"][int(tracer1[-1])]**2
```
carlosggarcia, line 888: This assumes you will have up to 10 lens tracers. But if you end up having 11, `tracer1[-1]` will be 1 instead of 10.

carlosggarcia (changes requested):
Looks good. A few minor comments apart from this one:
 - I think in `txpipe/theory.py` the EB, BE and BB Cells are not computed. Since you have the full covariance, including those, and that you also have those measurements, I think it'd be great to compute them for the fiducial cosmology. That would allow to do null tests too.

txpipe/twopoint_fourier.py
jpratmarti: I think it was needed before to do a cl guess but now it uses the workspace, so I removed it.

txpipe/twopoint_fourier.py
jpratmarti: I think that's OK now because it's only for the gaussian sims case where we have 5 bins only.

txpipe/covariance.py
```diff
@@ -788,12 +813,22 @@ def get_workspaces_dict(self, cl_sacc, masks_names):
...
 
-        w = {}
+        w = {'00': {}, '02': {}, '22': {}}
+        # Get the number of data points per Cell
+        dtype = cl_sacc.get_data_types()[0]
+        trs = cl_sacc.get_tracer_combinations()[0]
+        ell_eff, _ = cl_sacc.get_ell_cl(dtype, *trs)
+        n_ell = ell_eff.size
         for tr1, tr2 in cl_sacc.get_tracer_combinations():
+            # This assumes that the name of the tracers will be 'lens'or
```
joezuntz, line 823: I think it actually assumes that the name will include "lens" or "source", not actually *be* that.

txpipe/covariance.py
```diff
@@ -788,12 +813,22 @@ def get_workspaces_dict(self, cl_sacc, masks_names):
...
+            # 'source'
+            s1 = 0 if 'lens' in tr1 else 2
+            s2 = 0 if 'lens' in tr2 else 2
+            sk = ''.join(sorted(f'{s1}{s2}'))
             m1 = masks_names[tr1]
             m2 = masks_names[tr2]
             key = (m1, m2)
-            if (key in w) or (key[::-1] in w):
+            if (key in w[sk]) or (key[::-1] in w[sk]):
```
joezuntz, line 831: Not sure what I was doing here with the `key[::-1]` part.  If you know could you add a comment!

txpipe/twopoint.py
```diff
@@ -104,10 +104,14 @@ def run(self):
         for i, j, k in calcs:
             result = self.call_treecorr(i, j, k)
             results.append(result)
+            self.memory_report(f"DONE_TREECORR_{i}_{j}_{k}")
```
joezuntz, line 107: We can remove all the `memory_report` calls now.

txpipe/twopoint.py
```diff
@@ -317,50 +292,140 @@ def write_output(self, source_list, lens_list, meta, results):
 
         # Add the covariance.  There are several different jackknife approaches
         # available - see the treecorr docs
-        cov = treecorr.estimate_multi_cov(comb, self.config["var_method"])
+        print(comb[0])
```
joezuntz, line 295: and remove these extra print statements.

txpipe/twopoint.py
```diff
@@ -389,7 +454,17 @@ def call_treecorr(self, i, j, k):
         This is a wrapper for interaction with treecorr.
         """
         import sacc
-
+        import pickle
```
joezuntz, line 457: Can you add a `TODO` here for me to fix up the caching code?

txpipe/twopoint_fourier.py
```diff
@@ -142,6 +143,7 @@ def run(self):
         if self.rank == 0:
             nband = ell_bins.get_n_bands()
             ell_effs = ell_bins.get_effective_ells()
+            # this ell_effs variable is not used anywhere else, should we just remove it?
```
joezuntz, line 146: If it's not needed then yep, sure, please do remove it!

txpipe/twopoint_fourier.py
```diff
@@ -563,6 +559,14 @@ def compute_power_spectra(
                     0,
                     CEE,
                 ),
+                (
```
joezuntz, line 562: After you've made all your fixes could you run `black` on this file to tidy the layout?  Actually on all the changed files.

txpipe/twopoint_fourier.py
```diff
@@ -583,6 +587,7 @@ def compute_power_spectra(
 
         if self.config["analytic_noise"]:
             # we are going to subtract the noise afterwards
+            #if k == POS_POS:
```
joezuntz, line 590: remove

txpipe/twopoint_fourier.py
```diff
@@ -632,15 +638,22 @@ def window_pixel(ell, nside):
             y = f * x
             return np.exp(-(y**2) / 2)
 
+        # The binning information - effective (mid) ell values and
+        # the window information
+        ls = ell_bins.get_effective_ells()
+        
         c_beam = c / window_pixel(ls, pixel_scheme.nside) ** 2
-
+        #print("window_pixel(ls, pixel_scheme.nside) ** 2", window_pixel(ls, pixel_scheme.nside) ** 2)
```
joezuntz, line 646: remove old commented code

txpipe/utils/nmt_utils.py
```diff
@@ -72,14 +71,14 @@ def apply_window(self, b, c_ell):
         return (c_ell[ell] * weight).sum() / weight.sum()
 
     @classmethod
-    def from_binning_info(cls, ell_min, ell_max, n_ell, ell_spacing):
+    def from_binning_info(self, ell_min, ell_max, n_ell, ell_spacing):
```
joezuntz, line 74: This shouldn't be `self`, it should be `cls`.  That's because `@classmethod` makes it so this method isn't attached to a specific instance of the class, but the class itself.  So the first argument isn't an instance of `MyNmtBinFlat`, it's the class object `MyNmtBinFlat` itself.  That means that we can only call other class methods from this method.  And the standard is to make this clear by calling the first arg `cls`.

More info: https://stackoverflow.com/questions/12179271/meaning-of-classmethod-and-staticmethod-for-beginner

joezuntz (changes requested):

txpipe/covariance.py
```diff
@@ -692,22 +706,32 @@ def run(self):
...
-        # Would it be better to pass a path? Masks are not that heavy, so we
-        # might save some I/O overhead reading them at once here)
+        # For clustering, we follow twopoint_fourier.py:214-219
         with self.open_input("mask", wrapper=True) as f:
             mask = f.read_map("mask")
+            mask[mask == healpy.UNSEEN] = 0.0
             if self.rank == 0:
                 print("Loaded mask")
 
+        # Set any unseen pixels to zero weight.
```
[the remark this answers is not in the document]
jpratmarti: What do you mean here exactly? TODO what?

txpipe/covariance.py
```diff
@@ -387,18 +400,17 @@ def compute_covariance_block(
...
+    def get_angular_bins(self, cl_sacc):
+        # This function replicates `choose_ell_bins` in twopoint_fourier.py
+        # TODO: Move this to txpipe/utils/nmt_utils.py
+        from .utils.nmt_utils import MyNmtBin
+
+        ell_min = cl_sacc.metadata["binning/ell_min"]
+        ell_max = cl_sacc.metadata["binning/ell_max"]
+        ell_spacing = cl_sacc.metadata["binning/ell_spacing"]
+        n_ell = cl_sacc.metadata["binning/n_ell"]
+        edges = np.unique(np.geomspace(ell_min, ell_max, n_ell).astype(int))
```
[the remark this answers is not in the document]
jpratmarti: I copied that from here: https://github.com/LSSTDESC/TXPipe/blob/2c22aaf12d1396814eb4541b968ede6d9277e3b7/txpipe/utils/nmt_utils.py#L78, and it also matches I believe the function in twopoint_fourier. What alternative do you have in mind?

txpipe/covariance.py
```diff
@@ -730,6 +754,7 @@ def run(self):
 
         # Load NmtBin used for the Cells
         workspaces = self.get_workspaces_dict(cl_sacc, masks_names)
+        print('workspaces:', workspaces)
```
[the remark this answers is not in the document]
jpratmarti: Nope, removed.

txpipe/covariance.py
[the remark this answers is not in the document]
joezuntz: Just a comment saying 
```
# TODO: unify this code with the code in twopoint_fourier.py
```

txpipe/covariance.py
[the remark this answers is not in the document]
joezuntz: Okay, if it's what we're using already that's fine.  It's definitely okay for larger ell.

txpipe/covariance.py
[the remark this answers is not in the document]
jpratmarti: Done.

txpipe/covariance.py
jpratmarti: I think key[::-1] reverses the order so you were probably checking m1, m2 or m2, m1 combinations right? I added a comment.

txpipe/twopoint.py
jpratmarti: Done.

txpipe/twopoint.py
jpratmarti: Done.

carlosggarcia (approved):
Hi @jpratmarti. I'm happy with your changes now. I think you can merge when Joe is happy too.

txpipe/covariance.py
joezuntz: Thanks!

joezuntz (approved):

PR #247, opened 2022-05-05
Rail summarize interface

joezuntz:
This is currently work-in-progress.


eacharles (approved):
Looks good for now.   

We should probably revisit how we are doing the summarization and how we can use qp more transparently in the iteration / writing down the road, but let's talk about that at the collaboration meeting.

joezuntz:
Thanks! Yep, let's talk in Chicago.

sjoudaki:
Thanks - we accept this on behalf of the DESC PZ Working Group.

Issue #250, opened 2022-06-13
clearing up which stages are non-parallel


empEvil:
We need to go through and make sure that all stages that are non-parallel are set to `parallel=False`. In a similar fashion we need to get this be reflected in the documentation

empEvil:
Just as an update and reminder for us. We still need to figure out how to reflect this in the documentation!

joezuntz:
Finally dealt with in #382 !  The parallelism type is shown on the readthedocs stage listings under here: https://txpipe.readthedocs.io/en/latest/stages.html

PR #252, opened 2022-06-16
Label non-parallel stages as such

joezuntz:
Addresses issue #250 .

Also renames some classes which did not match their name attribute


empEvil (commented):
This is great. If I understand Ceci correctly it should now mean that if we try to `nprocess =2` for this it should warn us or stop it? Also more importantly how do we ensure the lack of parallelization is reflected in documentation, rather than by guess?

joezuntz:
It should complain if you try and run with nprocess=2, but yes, we should try and make the doc generator reflect it somehow.  And maybe write a few notes in docstring about what good choices are.

empEvil (approved):

empEvil:
we shpuld leave up the issue, for the documentations side

PR #256, opened 2022-07-22
Implemented TJPCov SSC 

carlosggarcia:
This implements the interface with TJPCov to compute the SSC. Now, the stage `TXFourierTJPCovariance` outputs not only `summary_statistics.sacc` but also `summary_statistics_gauss.sacc` and `summary_statistics_ssc.sacc` if they are requested. In order to compute both the Gaussian and the SSC terms, one has to define in the corresponding stage configuration in `config.yml` the following variable `cov_type = ['gauss', 'ssc']`.

The SSC implementation in TJPCov is at the moment under review (https://github.com/LSSTDESC/TJPCov/pull/34), so it might contain some bug.

### DIFF
--- a/examples/metadetect/config.yml
+++ b/examples/metadetect/config.yml
@@ -265,6 +265,8 @@ TXRealGaussianCovariance:
     nbins: 10
     pickled_wigner_transform: data/example/inputs/wigner.pkl
 
+TXFourierTJPCovariance:
+    cov_type: ['gauss', 'ssc']
 
 TXJackknifeCenters:
     npatch: 5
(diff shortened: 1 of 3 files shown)


carlosggarcia:
For a cosmological analysis you will always want both terms but if you are interested only on the SSC for some reason, it will be much faster if you just compute that instead of both the Gaussian and SSC parts. I've implemented the first suggestion (that it's one of 'gauss' or 'ssc')

carlosggarcia:
back to you, @joezuntz. Thanks!

joezuntz:
Looks great - I'll merge once the tests pass.

joezuntz (approved):

PR #264, opened 2022-08-31
Use RAIL installed with pip

joezuntz:
This removes the RAIL submodule and replaces it in installation with the new pypi versions created by the PZ group


eacharles (approved):
Overall it looks good.  My only question is if the "PZRailEstimateLensFromSource" and "PZRailEstimateSourceFromLens" could be avoided by using aliases to set the inputs of the following stages?

joezuntz:
Thanks - v good point! I've done that now and everything seems good, so merging.

PR #267, opened 2022-09-21
Add files via upload

jgpaine:


joezuntz:
Hi Johanna - I've had a chance to look at this. I think you might have put in the wrong file, since the ratios are still in this one? You can see the new files here: https://github.com/LSSTDESC/TXPipe/pull/267/files

Also it would be most useful to add the two new classes to the same existing file, txpipe/twopoint_plots.py  - happy to chat about the process for that.

joezuntz:
This is superseded by #273

PR #276, opened 2022-11-14
Update source_selector.py

jjeffe14:
fixing bug--replacing shear_data with data


chihway:
@empEvil if this looks good can you approve it?

PR #278, opened 2022-11-23
Tessore randoms

elvinpoole:
Added option to generate randoms using ntessore/healpix package (https://github.com/ntessore/healpix)

Faster than generating per pixel from quadrilateral (~0.3min vs 10min on one core) and will maintain matching to healpix pixel ids when using healpy tools

### DIFF
--- a/txpipe/random_cats.py
+++ b/txpipe/random_cats.py
@@ -9,6 +9,7 @@ from .data_types import (
 )
 from .utils import choose_pixelization, Splitter
 import numpy as np
+import healpix
 
 
 class TXRandomCat(PipelineStage):
@@ -33,6 +34,7 @@ class TXRandomCat(PipelineStage):
         "Mstar": 23.0,  # Schecther distribution Mstar parameter
         "alpha": -1.25,  # Schecther distribution alpha parameter
         "chunk_rows": 100_000,
+        "method":"quadrilateral", #method should be "quadrilateral" or "spherical_projection"
     }
 
     def run(self):
@@ -74,6 +76,9 @@ class TXRandomCat(PipelineStage):
         Mstar = self.config["Mstar"]
         alpha15 = 1.5 + self.config["alpha"]
         density_at_median = self.config["density"]
+        method = self.config["method"]
+        allowed_methods = ["quadrilateral","spherical_projection"]
+        assert method in allowed_methods
 
         # Work out the normalization of a Schechter distribution
         # with the given median depth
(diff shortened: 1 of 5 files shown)


joezuntz:
Looks like the tests pass now - is it ready for a review?

elvinpoole:
yes I think so

joezuntz:
Great - @empEvil volunteered at the hack time, so I've assigned him, thanks Eske!

txpipe/random_cats.py
```diff
@@ -168,7 +175,7 @@ def run(self):
             # Generate the random points in each pixel
             ndone = 0
 
-            nvertex = len(vertices)
+            nvertex = npix #number of verticies is the same as number of pixels
```
empEvil, line 178: Just trying to make sure I understand, is this change significant?

empEvil (commented):
Great idea, I have a few questions though. Does this break any of the standard examples? If so they should be fixed first! I would suggest that if it isn't too much work this gets made into a sub-class instead, and people can then call the subclass rather than the method to run it.

txpipe/random_cats.py
elvinpoole: no, it changes nothing 
len(vertices) == npix
We can switch it back if that's preferred

elvinpoole:
On the standard examples. I believe they all run as they did before. The default behaviour is currently still to use the old method

elvinpoole:
happy to add it as a sub-class if that makes more sense. I didn't originally as I didn't want us to end up with too many sub-classes for all the different ways of making randoms that might eventually be needed. e.g. depth varying vs uniform, generated from systematic weights, etc

empEvil:
I think this is something we should also get @joezuntz input on

txpipe/random_cats.py
```diff
@@ -231,7 +279,7 @@ def run(self):
                 )
 
                 # Save output to the generic non-binned output
-                index = bin_starts[j] + pix_starts[j, i]
+                index = bin_starts[j] + pix_starts[j, i] #this line was in the quadrilateral method but didn't appear to be doing anything...
```
joezuntz, line 282: Might be left over from an earlier version Feel free to remove!

joezuntz:
I'm happy to leave it as one class for now since there's only two methods here, but lets see how things go later as we develop. Maybe later it will make more sense to split.

joezuntz:
You could add a run of this pipeline to the CI. It looks like it would run after the metadetect one, so maybe put it in the same run as that?  You would just have to add the line to .github/workflows/ci.yml

empEvil (approved):
In that case I don't have any issues with us merging this.

elvinpoole:
ok great, I'll look into adding to the CI now

elvinpoole:
is the shifter on nersc updated to use the healpix module @joezuntz ?

joezuntz:
I've just started pulling now, it should take < 5 min.

joezuntz:
@elvinpoole  looks good - happy for me to merge?

PR #279, opened 2022-11-28
Move all theory calculations to new firecrown

joezuntz:
This PR:
- updates requirements to use the newest version of firecrown
- unifies the theory calculations in blinding and theory stages to a single function, theory_3x2pt, which itself calls firecrown

CosmoSIS is not actually needed to be installed here as far as I can tell, since we hook directly into the firecrown CCL objects instead of building a pipeline. We could change that later if we want a more complicated pipeline with IAs etc in the theory.

This is work-in-progress as I will need to update the docker images with the newest firecrown.


txpipe/utils/theory.py
```diff
@@ -51,111 +51,206 @@ def ccl_read_yaml(filename, **kwargs):
...
+                    continue
+                else:
+                    raise
+
+            # Read statistic information from the sacc file
+            stat.read(sacc_data)
+
+            # The first output is the data vector, and
+            # we keep the second which is the theory
+            _, theory[dt, t1, t2] = stat.compute(cosmo)
```
tilmantroester, line 194: I'd avoid using this low-level API (`stat.compute`) for now.
There's a firecrown PR in the works to add supports non-linear models (PT, halo model, etc). This changes the `cosmo` object here to a firecrown `Cosmology` object that carries around various CCL calculators along the CCL cosmology object. For this to work, the top level of the firecrown pipeline (the likelihood objects) now does some setup stuff too. TL;DR: this API will break soon and also lock you out of more sophisticated models in firecrown.

So I'd set up a dummy likelihood here,  use that to compute the theory predictions, and then extract the predictions.
There's a firecrown issue (https://github.com/LSSTDESC/firecrown/issues/195) to make this less awkward.

tilmantroester:
> We could change that later if we want a more complicated pipeline with IAs etc in the theory.

There's already support for various systematics in firecrown (incl. IA) and there will be significantly more, so falling back to cosmosis shouldn't be necessary.

joezuntz:
Hi @tilmantroester  - I've done a tidy of this, using the proper FireCrown 1.4 functionality. Could you re-review?

txpipe/utils/theory.py
joezuntz: Updated to use a dummy likelihood.

txpipe/utils/theory.py
```diff
@@ -51,111 +55,269 @@ def ccl_read_yaml(filename, **kwargs):
...
+
+    systematic_params = {
+        **make_bias_parameters(bias, sacc_data, cosmo)
+    }
+    # Apply the systematics parameters
+    likelihood.update(ParamsMap(systematic_params))
+
+    # Run the likelihood. As a by-product this fills in
+    # the predicted data vector
+    loglike = likelihood.compute_loglike(tools)
```
tilmantroester, line 198: There's now also `GaussFamily.compute_theory_vector`, which returns the same data as `predicted_data_vector` and which in the future hopefully won't require actual data or covariance in the sacc file.

As far as I can tell, the likelihood-level systematics were moved to the `ModellingTools` class, so using `compute_theory_vector` is fine.

I'm hoping to have a clean theory-only interface for firecrown at some point.

txpipe/utils/theory_model.py
```diff
@@ -0,0 +1,99 @@
...
+    # sources are saved in a dictionary since they will be used by one or more
+    # two-point function.
+
+    sources = {}
+
+    for tracer_name in sacc_data.tracers.keys():
+        if tracer_name.startswith("source"):
+            tr = wl.WeakLensing(sacc_tracer=tracer_name)
+        elif tracer_name.startswith("lens"):
+            bias_systematic = nc.LinearBiasSystematic(sacc_tracer=tracer_name)
```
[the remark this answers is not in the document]
joezuntz: I'd like to leave this for now in case we transition to a more complicated bias systematic in future - this would let us leave this model alone.

txpipe/utils/theory_model.py
```diff
@@ -0,0 +1,99 @@
...
+            tr = nc.NumberCounts(sacc_tracer=tracer_name, systematics=[bias_systematic])
+            i = int(tracer_name.split('_')[1])
+            tr.mag_bias = 0.0
```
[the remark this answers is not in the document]
joezuntz: Thanks - removing.

txpipe/theory.py
```diff
@@ -36,152 +37,19 @@ def run(self):
             matter_power_spectrum="halofit", Neff=3.046
         )
         print(cosmo)
-        s_theory = self.replace_with_theory_real(s, cosmo)
+        bias = self.config['galaxy_bias']
+        if bias == [0.0]:
+            bias = None
+        elif len(bias) == 1 and bias[0] < 0:
+            bias = -bias[0]
```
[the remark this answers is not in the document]
joezuntz: Added some comments on this. It's not intuitive at all but we couldn't think of a better approach.

txpipe/utils/theory.py
joezuntz: Thanks, I didn't know about this. I've switched to using it.

tilmantroester (approved):
LGTM!

PR #280, opened 2022-11-28
Fix theory plots

jgpaine:
Hi, I added the theory plot subclass, imported it, and added it to config.


joezuntz (approved):
Thanks! I've made some minor tweaks (there was additional old code that I could remove, left over from a while ago). Merging now!

PR #282, opened 2022-12-30
Use the pypi version of TJPCov

carlosggarcia:
I have removed the `submodules/TJPCov` dependency and used the pypi version of TJPCov. I have also modified a few imports that will be broken in tjpcov-0.1.2 (i.e. after merging https://github.com/LSSTDESC/TJPCov/pull/62). The new way is compatible with current version of TJPCov, too.

Edit: extra commits replacing `np.bool` by `bool` and `np.float` by `float` have been pushed to avoid the tests failing due to numpy deprecations. Would have it been better to change `np.float` to `np.float64`?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,9 +5,6 @@
 [submodule "WLMassMap"]
 	path = submodules/WLMassMap
 	url = https://github.com/LSSTDESC/WLMassMap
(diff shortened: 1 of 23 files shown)


joezuntz (approved):
Looks good!

PR #283, opened 2023-01-25
Add new version of the "Welcome" notebook

joezuntz:


nsevilla:
Typos:

Intro cell: 
* separate TxPipeand Jupyter
* Link to CosmoDC2 arXiv: https://arxiv.org/abs/1907.06530
* pipline --> pipeline

Running the pipeline:
* I would eliminate the first line of the second paragraph as it says the same thing as the first line of the cell. Leads to some confusion.
* Maybe some details or printout of the config file could be shown at this point (stages and inputs).

HDF5 catalogs:
* folow --> follow
* All subsequent code looks for things in 'output ' subdirectory, and actually I think it is 'outputs_metadetect', according to what was executed before.
* 'all these things main' --> 'all these things mean'
* My session was not able to execute completely cell [7]. Will try again as soon as Perlmutter is back.

Redshift PDFs:
* distibution --> distribution
* does not compute PDFs of individual source galaxies

Measurement in SACC files:
* argumnet --> argument

I get a lot of matplotlib warnings with (I think) default DESC setup.

PR #284, opened 2023-02-10
fixed mpi bugs in TwoPoint 

elvinpoole:
(1) calculateXi now run on all ranks

(2) Treecorr catalog objects are now created only on rank==0 along with an initial call of the treecorr cat.get_patches()
This was needed so that the patched catalog cache files are only saved by one rank, and don't cause conflicts

### DIFF
--- a/txpipe/twopoint.py
+++ b/txpipe/twopoint.py
@@ -642,39 +642,61 @@ class TXTwoPoint(PipelineStage):
         import treecorr
 
         # Load and calibrate the appropriate bin data
-        cat = treecorr.Catalog(
-            self.get_input("binned_shear_catalog"),
-            ext=f"/shear/bin_{i}",
-            g1_col="g1",
-            g2_col="g2",
-            ra_col="ra",
-            dec_col="dec",
-            w_col="weight",
-            ra_units="degree",
-            dec_units="degree",
-            patch_centers=self.get_input("patch_centers"),
-            save_patch_dir=self.get_patch_dir("binned_shear_catalog", i),
-            flip_g1=self.config["flip_g1"],
-            flip_g2=self.config["flip_g2"],
-        )
+        if self.rank == 0:
+            cat = treecorr.Catalog(
+                self.get_input("binned_shear_catalog"),
+                ext=f"/shear/bin_{i}",
+                g1_col="g1",
+                g2_col="g2",
+                ra_col="ra",
+                dec_col="dec",
+                w_col="weight",
+                ra_units="degree",
+                dec_units="degree",
+                patch_centers=self.get_input("patch_centers"),
+                save_patch_dir=self.get_patch_dir("binned_shear_catalog", i),
+                flip_g1=self.config["flip_g1"],
+                flip_g2=self.config["flip_g2"],
+            )
+
+            #run get_patches on rank==0 only
+            if cat._patches is None:
+                cat.get_patches()
+        else:
+            cat = None 
+
+        if self.comm is not None:
+            cat = self.comm.bcast(cat, root=0)
 
         return cat
+
 
     def get_lens_catalog(self, i):
         import treecorr
 
         # Load and calibrate the appropriate bin data
-        cat = treecorr.Catalog(
-            self.get_input("binned_lens_catalog"),
-            ext=f"/lens/bin_{i}",
-            ra_col="ra",
-            dec_col="dec",
-            w_col="weight",
-            ra_units="degree",
-            dec_units="degree",
-            patch_centers=self.get_input("patch_centers"),
-            save_patch_dir=self.get_patch_dir("binned_lens_catalog", i),
-        )
+        if self.rank == 0:
+            cat = treecorr.Catalog(
+                self.get_input("binned_lens_catalog"),
+                ext=f"/lens/bin_{i}",
+                ra_col="ra",
+                dec_col="dec",
+                w_col="weight",
+                ra_units="degree",
+                dec_units="degree",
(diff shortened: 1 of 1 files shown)


txpipe/twopoint.py
```diff
@@ -759,9 +792,9 @@ def calculate_shear_pos(self, i, j):
...
             rg = None
 
+        ng.calculateXi(rg=rg)
+        t2 = perf_counter()
         if self.rank == 0:
-            ng.calculateXi(rg=rg)
-            t2 = perf_counter()
             print(f"Processing took {t2 - t1:.1f} seconds")
 
         return ng
```
empEvil, line 800: seems to me as you are making each core calculate Xi now rather than only rank 0? Isn't Xi calculated based on the recombined results from all the processes? Is this intended?

txpipe/twopoint.py
```diff
@@ -818,9 +851,9 @@ def calculate_pos_pos(self, i, j):
             rn = treecorr.NNCorrelation(self.config)
             rn.process(rancat_i, cat_j, comm=self.comm, low_mem=self.config["low_mem"])
 
+        t2 = perf_counter()
+        nn.calculateXi(rr, dr=nr, rd=rn)
         if self.rank == 0:
-            t2 = perf_counter()
-            nn.calculateXi(rr, dr=nr, rd=rn)
             print(f"Processing took {t2 - t1:.1f} seconds")
```
empEvil, line 858: See comment above

txpipe/twopoint.py
elvinpoole: yes its intended. The issue was ng.xi is needed by all processes later in `write_output`. So if you only run calculateXi on rank==0 there will be an error on the other ranks. since calculateXi is so light it was easier to just have all processes run it

elvinpoole: alternatively we could compute ng.xi on rank==0 and bcast to the other ranks

empEvil: That is what confuses me cause if you look at line 414 in the write_ouput function other than rank==0 should terminate before they need the term estimated from calculateXi? I think at least that was the intend? @joezuntz any comments on this?

elvinpoole: I think the error comes before 414 in write output. in add_data_points

Here is the error 
```
Traceback (most recent call last):
 File "/opt/conda/lib/python3.9/runpy.py", line 197, in _run_module_as_main
Using new TreeCorr 4.3 or above
  return _run_code(code, main_globals, None,
 File "/opt/conda/lib/python3.9/runpy.py", line 87, in _run_code
  exec(code, run_globals)
 File "/global/cscratch1/sd/jelvinpo/DESC/TXPipe/txpipe/__main__.py", line 5, in <module>
  PipelineStage.main()
 File "/opt/conda/lib/python3.9/site-packages/ceci/stage.py", line 469, in main
  stage.execute(args)
 File "/opt/conda/lib/python3.9/site-packages/ceci/stage.py", line 604, in execute
  stage.run()
 File "/global/cscratch1/sd/jelvinpo/DESC/TXPipe/txpipe/twopoint.py", line 126, in run
  self.write_output(source_list, lens_list, meta, results)
 File "/global/cscratch1/sd/jelvinpo/DESC/TXPipe/txpipe/twopoint.py", line 419, in write_output
  self.add_data_points(S, results)
 File "/global/cscratch1/sd/jelvinpo/DESC/TXPipe/txpipe/twopoint.py", line 308, in add_data_points
  xi = d.object.xi
AttributeError: 'NNCorrelation' object has no attribute 'xi'
```

elvinpoole: The actual error comes from line 301

empEvil: You are correct it happens at 412 in `write_output`, when it calls `add_data_points`, question then from me is is there a reason that each cpu adds data points, or should that be initiated by just the rank==0? as far as I can see the `add_data_points` calls TreeCorr with a mpi comm object already. I think your solution solves the error, I will try and test it later. But it does add confusion for me to see how we are parallelizing these steps of the code. TreeCorr has built in mpi, which we are utilizing but are we also parallelizing the calls of TreeCorr themself?

elvinpoole: yea its a good point. only one rank really needs to construct the sacc file

empEvil (approved):
We discussed this, and I think I am fine merging this for now

joezuntz:
@elvinpoole did you resolve the questions we had about the patch making?

elvinpoole:
> @elvinpoole did you resolve the questions we had about the patch making?

I think I figured out what is happening, but haven't fixed it yet. I think when there are empty patches (as there are in the metadetect example) the txpipe prepare_patches will remove the empty patch and rename the other files to be 0 to N. Treecorr doesn't know about this so thinks there are missing files and  it remakes the patches

elvinpoole:
My fix does prevent the mpi error in the treecorr patches function but what we actually want is to stop treecorr from making new patches

elvinpoole:
I moved the get_patches call option to a separate method. This now works as intended (i.e. does the same thing as main branch but allows MPI with empty patches)

Details: It seems in general the issue is the treecorr get_patches is unable to load even its own patch files when there is an empty patch (it will still save/rename the patch files in the cache, but every time the catalog is called it just re-makes them, rather than having to remember which patch files it deleted)

I tried a few ways to fix this (including making empty patch files) but couldn't figure it out using only TXpipe so maybe we should open a treecorr issue 

For now, I would vote to leave the behaviour as it is here (re-making patch files for each 2pt function - only when empty patches are present) given empty patches *should* be an unusual use-case

PR #286, opened 2023-03-22
Subsample randoms savecat

elvinpoole:
- added stage to randomly subsample the binned randoms catalog and save as separate file
- added config option to twopoint to use the subsampled catalog for the RR clustering term

tested on cosmodc2 example. Speeds up the auto correlations by 30% and cross correlations by 50% (will do better for larger random/data ratios) and negligible change in correlation function

### DIFF
--- a/txpipe/random_cats.py
+++ b/txpipe/random_cats.py
@@ -26,6 +26,7 @@ class TXRandomCat(PipelineStage):
     outputs = [
         ("random_cats", RandomsCatalog),
         ("binned_random_catalog", RandomsCatalog),
+        ("binned_random_catalog_sub", RandomsCatalog),
     ]
     config_options = {
         "density": 100.0,  # number per square arcmin at median depth depth.  Not sure if this is right.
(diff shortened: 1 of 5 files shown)


elvinpoole:
currently needs 
```
inputs:
    binned_random_catalog_sub: filename
```
to run, even when not using this feature, unless there is a way to add optional inputs to a stage?

txpipe/random_cats.py
```diff
@@ -294,11 +297,171 @@ def run(self):
...
+                            )
+
+            batch2.finish()
+
+        if self.comm is not None:
+            self.comm.Barrier()
+        binned_output_sub.close()
+
+
+class TXSubsampleRandoms(PipelineStage):
```
joezuntz, line 380: I think this class is no longer used?

elvinpoole: that's right. I left it in because I figured we might be adding new random catalog stages in the future (eg. SSI) and this way we could use TXSubsampleRandoms with those. Happy to remove it if we just want to insist any random catalog stage has to produce binned_random_catalog_sub as well

joezuntz: Good point! Let’s leave it in.

joezuntz (approved):

PR #289, opened 2023-04-12
Fracdet

elvinpoole:
Added a stage to track fractional coverage of pixel when making mask, TXSimpleMaskFrac
Requires a healsparse map with higher-resolution than requested mask to be specified. Then uses healsparse fracdet calculation (fraction of valid sub-pixels)

TXRandomCat uses this fractional coverage when generating randoms


elvinpoole:
I have tested this on 1deg2 metadetect example. I cut the supreme Nexposures map down to the same 1deg2 patch and used this input map to define fracdet. This could be added to example data if we want

elvinpoole:
checked for places downstream where binary mask is expected. Only one needed changing was the twopoint pixel estimator. We now weight each pixel by its fracdet. Density maps themselves already accounted for non-binary mask

elvinpoole:
Yes I think so

txpipe/masks.py
```diff
@@ -61,7 +69,59 @@ def run(self):
...
+    name = "TXSimpleMaskFrac"
+    parallel = False
+    # make a mask from the auxiliary maps
+    inputs = [("aux_lens_maps", MapsFile),
+    ]
+    outputs = [("mask", MapsFile)]
+    config_options = {
+        "depth_cut": 23.5,
+        "bright_object_max": 10.0,
+        "supreme_map_file": None,
```
joezuntz, line 94: I can't recall but I think it might be better to set the default to an empty string instead of None so that argparse knows what kind of type to expect.

txpipe/random_cats.py
```diff
@@ -93,8 +98,10 @@ def run(self):
         x = 10.0 ** (0.4 * (Mstar - depth))
         density = phi_star * scipy.special.gammaincc(alpha15, x)
 
-        # Pixel geometry - area in arcmin^2
-        pix_area = scheme.pixel_area(degrees=True) * 60.0 * 60.0
+        # Pixel geometry - area of a single pixel in arcmin^2
+        full_pix_area = scheme.pixel_area(degrees=True) * 60.0 * 60.0
+        pix_area = frac*full_pix_area
```
joezuntz, line 103: Maybe add a TODO here to use the higher resolution healsparse map to populate the randoms? Or does that not matter once you're at high enough res anyway?

joezuntz (changes requested):
Very minor changes!

txpipe/random_cats.py
elvinpoole: yea good idea. Our mask should be at high enough resolution that it wont matter, but it would be nice to have. Would require using the high-res "supreme" map as input rather than the mask

elvinpoole: as for this PR, this just means adding a comment in the code to say this is a TO DO Item?

joezuntz: A TODO is fine for now. Let's see if we need this.

txpipe/masks.py
joezuntz: Wait, think I confused myself on this - is this a required parameter or an optional one?  If it's mandatory you can set the default to `str`

elvinpoole: yes its required. I've now set it to str

txpipe/random_cats.py
elvinpoole: ok I've added that

joezuntz:
Thanks! If you don't need the branch any more for other stuff you can delete it now.

PR #293, opened 2023-05-30
Basic weights pipeline

elvinpoole:
Basic linear weights pipeline

Base stage `TXLSSweights` can be inherited by each weights method

`TXLSSweightsUnit`
Applies weight=1 to all objects

`TXLSSweightsSimReg`
Tests galaxy density vs Survey properties
Corrects for significant trends with linear regression


txpipe/auxiliary_maps.py
elvinpoole, line 1: TXAuxiliaryLensMaps now gets pixel info from config file rather than lens map. This prevents the circular module issue aux_lens_map -> mask -> LSSweights -> lens_map -> aux_lens_map

txpipe/maps.py
elvinpoole, line 1: if any aliases have been defined in the config to a maps stage, this can't be saved to the h5 attributes because it seems an attr can't be a dictionary. This edit just skips the alias and saves everything else. this is useful if you want to skip the weights stage using aliases

txpipe/lssweights.py
```diff
@@ -0,0 +1,1103 @@
...
+			mask = map_file.read_map("mask")
+
+		#load the ra and dec of this lens bins
+		with self.open_input("binned_lens_catalog_unweighted", wrapper=False) as f:
+			ra = f[f"lens/bin_{tomobin}/ra"][:]
+			dec = f[f"lens/bin_{tomobin}/dec"][:]
+			weight = f[f"lens/bin_{tomobin}/weight"][:]
+
+		#pixel ID for each lens galaxy
+		obj_pix = hp.ang2pix(nside,ra,dec,lonlat=True, nest=True)
```
joezuntz, line 147: This seems to assume nest=True but the bit below checks. Unless that's two different data sources, healsparse vs internal maps?

txpipe/lssweights.py
```diff
@@ -0,0 +1,1103 @@
...
+		pixel = pixels_bincount[Ncounts_bincount != 0.]
+		Ncounts = Ncounts_bincount[Ncounts_bincount != 0.]
+
+		if nest:
+			maskpix = np.where(mask!=hp.UNSEEN)[0]
+		else:
+			maskpix = hp.ring2nest(nside, np.where(mask!=hp.UNSEEN)[0])
+
+		#fractional coverage
+		frac = hsp.HealSparseMap.make_empty(32, nside, dtype=np.float64)
```
joezuntz, line 161: Will this always be 32?  If so maybe make it a constant at the top of the file. If not then an option.

txpipe/lssweights.py
```diff
@@ -0,0 +1,1103 @@
...
+		frac.update_values_pix(maskpix, mask[np.where(mask!=hp.UNSEEN)[0]])
+
+		deltag = hsp.HealSparseMap.make_empty(32, nside, dtype=np.float64)
+		deltag.update_values_pix(maskpix, 0.0)
```
joezuntz, line 165: This looks like you set the `maskpix` pixels of deltag  to zero and then a few lines later use `deltag[maskpix]` - wouldn't that just be zero?

txpipe/lssweights.py
```diff
@@ -0,0 +1,1103 @@
...
+		print(f"Found {nsys} total systematic maps")
+
+		with self.open_input("mask", wrapper=True) as map_file:
+			mask = map_file.read_map("mask")
+			mask_nest = map_file.read_map_info("mask")["nest"]
+			mask = hsp.HealSparseMap(
+				nside_coverage=32, 
+				healpix_map=(mask==hp.UNSEEN).astype('int'), 
+				nest=mask_nest, sentinel=0)
+			nside = map_file.read_map_info("mask")["nside"]
```
joezuntz, line 196: again can tidy to avoid calling this twice.

txpipe/lssweights.py
```diff
@@ -0,0 +1,1103 @@
...
+		from . import lsstools
+
+		s = time.time()
+
+		nsysbins = self.config["nbin"]
+		f = 0.5 * self.config["outlier_fraction"]
+		percentiles = np.linspace(f, 1 - f, nsysbins + 1)
+
+		with self.open_input("mask", wrapper=True) as map_file:
+			nside = map_file.read_map_info("mask")["nside"]
```
joezuntz, line 248: again here.

txpipe/lssweights.py
```diff
@@ -0,0 +1,1103 @@
...
+
+			#apply mask
+			sys_maps.append(sys_map)
+
+		f = time.time()
+		print("load_and_mask_sp_maps took {0}s".format(f-s))
+
+		return np.array(sys_maps), np.array(sys_names)
+
+	def calc_1d_density(self, tomobin, Fmap=None):
```
joezuntz, line 218: Could you give Fmap a more descriptive name?

txpipe/lssweights.py
```diff
@@ -0,0 +1,1103 @@
...
+		Fmap_list: list of Healsparse maps
+			list of the F(s) maps. one for each tomo bin
+			F(s) = 1/weight
+
+		"""
+		import healpy as hp
+		import matplotlib.pyplot as plt 
+
+		#### save the weights maps
+		map_output_file = self.open_output("lss_weight_maps", wrapper=True)
```
joezuntz, line 321: Need to close this, or use a `with` statement.

txpipe/lssweights.py
```diff
@@ -0,0 +1,1103 @@
...
+		for ibin in range(self.Ntomo):
+			#get weight per lens object
+			subgroup = binned_output[f"lens/bin_{ibin}/"]
+			ra = subgroup["ra"][:]
+			dec = subgroup["dec"][:]
+			pix = hp.ang2pix(self.pixel_metadata['nside'],ra,dec,lonlat=True,nest=True) #can switch to using txpipe tools for this?
+			obj_weight = 1./Fmap_list[ibin][pix]
+			obj_weight[obj_weight==1./hp.UNSEEN] = 0.0
+
+			subgroup["weight"][...] *= obj_weight
```
joezuntz, line 347: I think this should be a colon not ellipsis?

txpipe/lssweights.py
```diff
@@ -0,0 +1,1103 @@
...
+		#### save the tomography file
+		tomo_output = self.open_output("lens_tomography_catalog", parallel=True)
+		with self.open_input("lens_tomography_catalog_unweighted") as tomo_input:
+			tomo_output.copy(tomo_input["tomography"],"tomography")
+
+		lens_weight = tomo_output['tomography/lens_weight'][...]
+		for ibin in range(self.Ntomo):
+			binned_subgroup = binned_output[f"lens/bin_{ibin}/"]
+			mask = (tomo_output['tomography/bin'][...] == ibin)
+			lens_weight[mask] *= binned_subgroup["weight"][...]
```
joezuntz, line 371: I think this should be a colon instead of an ellipsis?

txpipe/lssweights.py
```diff
@@ -0,0 +1,1103 @@
...
+		density_correlation.save_to_hdf5(dens_corr_file_name)
+		weighted_dens_corr_file_name = output_dir.path_for_file(f"weighted_density_correlation_lens{ibin}.hdf5")
+		weighted_density_correlation.save_to_hdf5(weighted_dens_corr_file_name)
+
+		#save map names, coefficients and chi2 from the fit into an hdf5 file
+		fit_summary_file_name = output_dir.path_for_file(f"fit_summary_lens{ibin}.hdf5")
+		fit_summary_file = h5py.File(fit_summary_file_name, 'w')
+		fit_summary_file.file.create_group("fit_summary")
+		if len(fit_output['sig_map_index']) != 0:
+			fit_summary_file["fit_summary"].create_dataset('fitted_map_id', data=fit_output['sig_map_index'])
```
joezuntz, line 416: can tidy this using `summary_group = fit_summary_file["fit_summary"]` then reusing `summary_group`.

txpipe/lssweights.py
```diff
@@ -0,0 +1,1103 @@
...
+			if fit_output['coeff_cov'] is not None:
+				fit_summary_file["fit_summary"].create_dataset('coeff_cov', data=fit_output['coeff_cov'] )
+			for model in density_correlation.chi2.keys():
+				fit_summary_file["fit_summary"].create_dataset(f'chi2_{model}', data=np.array(list(density_correlation.chi2[model].items())).T )
+			fit_summary_file.close()
+
+		#plot the un-weighted and weighted chi2 distribution
+		filepath = output_dir.path_for_file(f"chi2_hist_lens{ibin}.png")
+		if "pvalue_threshold" in self.config.keys():
+			chi2_threshold = scipy.stats.chi2(self.config["nbin"]).isf(0.05)
```
joezuntz, line 430: shouldn't this 0.05 be `pvalue_threshold`?

txpipe/lssweights.py
```diff
@@ -0,0 +1,1103 @@
...
+				sys_obj_j = sys_map_j[obj_pix]
+				edgesj = density_correlation.get_edges(jmap)
+				maskj = np.where(density_correlation.map_index.astype('int') == jmap)[0]
+				startj = maskj[0]
+				finishj = maskj[-1]+1
+
+				n2d_pair,_,_ = np.histogram2d(sys_obj_i,sys_obj_j,bins=(edgesi,edgesj), weights=weight)
+				n2d[imap,jmap] = n2d_pair
+
+				#TO DO: CHECK I GOT THESE THE RIGHT WAY AROUND!!!!
```
joezuntz, line 591: Is this confirmed now?

txpipe/lssweights.py
```diff
@@ -0,0 +1,1103 @@
...
+
+
+class TXLSSweightsUnit(TXLSSweights):
+	"""
+	Class assigns weight=1 to all lens objects
+
+	"""
+	name = "TXLSSweightsUnit"
+	parallel = False
+
```
joezuntz, line 1045: I think you can override the inputs and outputs here since many of them are not made.  You might not even want this to descend from `TXLSSWeights` if it doesn't share much behaviour?

joezuntz (changes requested):

examples/weights_lss/pipeline.yml
```diff
@@ -0,0 +1,65 @@
...
+    - name: TXSimpleMaskFrac       # combine maps to make a simple mask
+    - name: TXLensNoiseMaps      # Compute lens noise using half-splits
+    - name: TXDensityMaps        # turn mask and ngal maps into overdensity maps
+
+    - name: TXLSSweightsSimReg   #compute LSS weights using linear regression 
+
+
+# modules and packages to import that have pipeline
+# stages defined in them
+modules: txpipe rail.stages
```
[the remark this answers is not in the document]
elvinpoole: I removed this example because it was redundant since adding the DES-y1 example

txpipe/lssweights.py
```diff
@@ -0,0 +1,1103 @@
...
+	outputs = [
+		("lss_weight_summary", FileCollection), #output files and summary statistics will go here
+		("lss_weight_maps", MapsFile), #the systematic weight maps to be applied to the lens galaxies
+		("binned_lens_catalog", HDFFile), #the lens catalog with weights added
+		("lens_tomography_catalog", HDFFile), #the tomography file with weights added
+	]
+
+	config_options = {
+		"supreme_path_root": "/global/cscratch1/sd/erykoff/dc2_dr6/supreme/supreme_dc2_dr6d_v2",
```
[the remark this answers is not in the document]
elvinpoole: I think this is an old path from cori, I copied it from TXMapCorrelations default input, but its not used in ny example Maybe we should just have a blank string as default?

txpipe/lssweights.py
elvinpoole: This line gets the pixels in which the each galaxy is positioned and is fixed to nest format
Below is getting the mask pixels. Currently it is possible to ingest an external mask in ring format, so the below checks will convert this to nest

txpipe/lssweights.py
elvinpoole: added nside_coverage as a config option

txpipe/lssweights.py
elvinpoole: so first I set the good pixels in the mask to 0, then fill in the pixels that actually contain galaxies with the line below `deltag.update_values_pix( pixel, Ncounts )` 
so when we use `deltag[maskpix]` only the empty pixels will be 0

txpipe/lssweights.py
elvinpoole: changed to mean_density_map

txpipe/lssweights.py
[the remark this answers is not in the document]
joezuntz: Yes, good idea.

txpipe/lssweights.py
joezuntz: There's a few other places where 32 is hard-coded - just checking if they should also be set from the config option?

txpipe/lssweights.py
joezuntz: Thanks!

txpipe/lssweights.py
joezuntz: Have you pushed this change? It's still called Fmap on github.

txpipe/lssweights.py
elvinpoole: confirmed

txpipe/lssweights.py
elvinpoole: The `save_weights` method from `TXLSSWeights` still gets used which uses the same inputs and output, so I think we can keep this as is?

examples/weights_lss/config.yml
```diff
@@ -0,0 +1,396 @@
...
+    # catalog data, and tells them how many rows to read
+    # at once
+    chunk_rows: 100000
+    # These mapping options are also read by a range of stages
+    pixelization: healpix
+    nside: 256
+    sparse: True  # Generate sparse maps - faster if using small areas
+
+TXLSSweightsSimReg:
+    supreme_path_root: txpipe_utils/fake_sp_maps/maps_gauss/
```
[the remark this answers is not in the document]
elvinpoole: I ended up removing this example because it was redundant after adding the DES-y1 example

txpipe/lsstools/lsstools.py
```diff
@@ -0,0 +1,205 @@
...
+			for map_index in self.mapnames:
+				select_map = (self.map_index == map_index)
+				ndens = self.ndens[select_map]
+				covmat = self.get_covmat_singlemap(map_index)
+				chi2_map = self.calc_chi2(ndens, covmat, model[select_map])
+				self.chi2[model_name][map_index] = chi2_map
+
+
+	@staticmethod
+	def calc_chi2(y, err, yfit , v = False):
```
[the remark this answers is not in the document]
elvinpoole: moved to `utils/fitting.py`

txpipe/lssweights.py
elvinpoole: yes, sorry i hadn't pushed a few changes, should be there now

txpipe/lssweights.py
elvinpoole: yes, sorry i hadn't pushed a few changes, should be there now

txpipe/lssweights.py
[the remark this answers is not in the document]
elvinpoole: done

txpipe/maps.py
```diff
@@ -120,7 +120,20 @@ def save_maps(self, pixel_scheme, maps):
         # add a maps section to each
         for output_file in output_files.values():
             output_file.file.create_group("maps")
-            output_file.file["maps"].attrs.update(self.config)
+            
+            #this is a bit of a hack, but if you have set an alias
```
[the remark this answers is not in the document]
elvinpoole: is this something that would have to change on the ceci side? or is there a way to save nested metadata already?

joezuntz (approved):

PR #294, opened 2023-06-21
Updates to cope with the RAIL renaming

joezuntz:


eacharles (approved):
This is all fine.  But we might want to chat a bit about how we deal with finding all the rail stuff.  I put a bit of introspection functionality into rail, so there might be easier ways to find the stages.   
Also you probably don’t need all of rail[dev].

PR #297, opened 2023-07-04
Add script to find ancestors of a stage or output

joezuntz:


joezuntz:
Moved this to ceci: https://github.com/LSSTDESC/ceci/pull/97

PR #299, opened 2023-07-13
Config for making a 20 deg^2 sample and some fixes needed

joezuntz:


joezuntz:
I'm being a bit bad and merging this before waiting for review because the CLMM people have a retreat and I have some fixes for them.

PR #317, opened 2023-08-01
Simplify tomography files and mappers

joezuntz:
Tomography files unnecessarily distinguish "source_bin" and "lens_bin" columns in files, making it harder to use them generically. 

This changes this so that all tomo files have a set of common columns.  At the same time, I removed the old TXMainMaps class, which assumed identical lens and source catalogs, because it would not work straightforwardly with this change (because both lens and source bins are now called "bin"), and rewrote the Mapper class as ShearMapper and LensMapper (it was never being used to do both at once outside TXMainMaps).

As a side point this finally does the source number count maps.

This was making the code more complicated than it needed to be when writing code to use QP and do truth photo-z summations.


joezuntz:
@empEvil did you have a chance to look at this?

empEvil:
No not yet, I happened to get sick :(

txpipe/mapping/basic_maps.py
```diff
@@ -190,15 +134,114 @@ def finalize(self, comm=None):
...
+
+        return pixel, count, g1, g2, var_g1, var_g2, source_weight, var_e
+
+
+class LensMapper:
+    def __init__(
+        self,
+        pixel_scheme,
+        bins,
+        do_lens=True,
```
empEvil, line 154: Does this Bool ever get called?

txpipe/noise_maps.py
```diff
@@ -619,6 +621,7 @@ def run(self):
             # The top section has the metadata in it
             group = outfile.file.create_group("maps")
             group.attrs["nbin_source"] = nbin_source
```
empEvil, line 623: is this line not redundant because of the line below?

txpipe/mapping/basic_maps.py
joezuntz: Good point - removed.

txpipe/noise_maps.py
```diff
@@ -447,6 +447,8 @@ def run(self):
         from jax import random, jit, device_get, device_put
         from .utils import choose_pixelization
 
+        raise ValueError("This code needs rewriting because source_bin and lens_bin now have the same name in the tomo files.")
```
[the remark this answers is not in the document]
joezuntz: This is the JAX version which nobody is currently using (it's more a proof-of-concept), so hopefully it shouldn't hurt anyone immediately. It does need rewriting but I was going to punt on that for a bit.  We can't leave it as a warning because it will just lead to a crash.

txpipe/noise_maps.py
joezuntz: I left a few things like this in as a backup in case there was old code that needed it.

joezuntz: I'll put in a TODO to fix it.

joezuntz:
Thanks for the review! As far as I can tell I've dealt with all the places where there's a name confusion. I'm sure we'll hit something else eventually!

joezuntz:
Could you approve if you're happy?

empEvil (approved):

joezuntz:
Thanks!!!

PR #319, opened 2023-08-03
Switch to using QP for n(z) information throughout

joezuntz:
This makes extensive changes to use QP objects for n(z) information throughout the code.

This PR includes #317 so should only be considered after that.

it also depends on QP PR [183](https://github.com/LSSTDESC/qp/pull/183).


txpipe/rail/summarize.py
```diff
@@ -121,105 +122,66 @@ def run(self):
...
-            if self.config["zmax"] > 0:
-                ax.set_xlim(0, self.config["zmax"])
-            ax.set_ylim(0, None)
-            ax.set_xlabel("z")
-            ax.set_ylabel("n(z)")
+
+        zmax = self.config["zmax"]
+        nz = self.config["nz"]
+        z = np.linspace(0, zmax, nz)
+        print(z)
```
chihway, line 165: remove this? or add info on what is being printed

txpipe/rail/summarize.py
```diff
@@ -1,20 +1,20 @@
...
     name = "PZRailSummarize"
     inputs = [
-        ("tomography_catalog", HDFFile),
+        ("tomography_catalog", TomographyCatalog),
         ("photometry_catalog", HDFFile),
         ("model", PickleFile),
 
     ]
     outputs = [
         # TODO: Change to using QP files throughout
```
chihway, line 15: remove?

txpipe/photoz_stack.py
```diff
@@ -1,496 +1,282 @@
...
 
-        # Save again but for the number of bins in the tomography catalog
-        with self.open_input("shear_tomography_catalog") as tomo_file:
-            nbin_source = tomo_file["tomography"].attrs["nbin"]
 
-        return z, nbin_source
 
 
-class TXPhotozLensStack(TXPhotozSourceStack):
+class TXTruePhotozStack(PipelineStage):
```
chihway, line 128: is there a version where we are using e.g. a z_mc value? that would be essentially the same as this class but histograming a different column (though I guess qp does not have that option now, and maybe not worth having...?)

examples/lensfit/pipeline.yml
```diff
@@ -95,6 +100,7 @@ inputs:
     # This file comes with the code
     fiducial_cosmology: data/fiducial_cosmology.yml
     flow: data/example/inputs/example_flow.pkl
+    none: none
```
chihway, line 103: just curious, what is this?!

txpipe/data_types/types.py
```diff
@@ -469,6 +438,136 @@ def to_ccl(self, **kwargs):
...
+        import tables_io
+        if self._metadata is not None:
+            return self._metadata
+        meta = tables_io.io.readHdf5GroupToDict(self.file["meta"])
+        self._metadata = meta
+        return meta
+
+class QPPDFFile(QPBaseFile):
+    def iterate(self, chunk_rows, rank=0, size=1):
+        import qp
```
chihway, line 459: dumb questions, why is this better to import in the functions?

.github/workflows/ci.yml
```diff
@@ -272,7 +272,6 @@ jobs:
         ceci --dry-run examples/2.2i/pipeline.yml
         ceci --dry-run examples/buzzard/pipeline.yml
         ceci --dry-run examples/cosmodc2/pipeline.yml
-        ceci --dry-run examples/dr1b/pipeline.yml
```
chihway, line 275: why did this get dropped?

examples/skysim/config.yml
```diff
@@ -24,6 +24,27 @@ TXIngestRedmagic:
 TXSourceTrueNumberDensity:
     nz: 301
     zmax: 3.0
+    name: TXSourceTrueNumberDensity
```
chihway, line 27: this is throughout, and Joe said it's a bug, might be a different PR tho

txpipe/data_types/types.py
```diff
@@ -469,6 +438,136 @@ def to_ccl(self, **kwargs):
...
+        if pdf_name == "interp":
+            z = meta["xvals"][:]
+        elif pdf_name == "hist":
+            z = meta["bins"][:]
+        else:
+            raise ValueError(f"TXPipe cannot understand pdf_name in QP file: {pdf_name}")
+        return z.squeeze()
+    
+    
+class QPMultiFile(HDFFile):
```
chihway, line 544: some comment explaining what this class is

chihway:
Ok @joezuntz I've done a pass, feel free to take or leave whatever I commented on!

txpipe/photoz_stack.py
joezuntz: I'm adding a config option to let people choose a different column here.  Using QP to generate draws would have to be from a different class though as the input would need to be a a QP file instead of a plain catalog. We can do that if needed though.

examples/lensfit/pipeline.yml
joezuntz: Adding a comment!

txpipe/data_types/types.py
joezuntz: You mean inside the function instead of at top level? That's to avoid loading QP until it's actually needed. It takes a while to load as it has so many dependencies. Also I avoid importing stuff until runtime throughout TXPipe for the same reason, and so that the launcher env can be simpler.

.github/workflows/ci.yml
joezuntz: It was just so old at this point, and superseded by 2.2i.

txpipe/rail/summarize.py
joezuntz: Done.

examples/skysim/config.yml
joezuntz: Yep - needs a fix in ceci!

joezuntz: Made an issue here: https://github.com/LSSTDESC/ceci/issues/100

txpipe/data_types/types.py
joezuntz: done!

joezuntz:
Thanks for reviewing @chihway ! This is merged now.

PR #320, opened 2023-08-23
Stage to make source maps from binned shear catalog instead of doing calibration internally

joezuntz:
This PR depends on #317 which contains almost all its contents, so should be looked at only after that is merged.

It replaces the source mapper with a dask-based version which is much simpler and doesn't use any complicated classes.


empEvil (commented):
Looks fine to me, but I am no expert on DASK so maybe get someone else to help if the DASK calls are correct? or can you link me to some details?

PR #324, opened 2023-10-05
Systematics taustats

yomori:
This PR implements tau stats in psf_diagnostic.py


yomori:
Cross-checked with published DES-Y3 results and get reasonable agreement, but Metacal responses have to applied manually since the Metacal implemented in TXPipe does things differently than how it was actually done in DES-Y3.

txpipe/psf_diagnostics.py
```diff
@@ -172,6 +172,415 @@ def plot_histogram(self, function, output_name, xlabel, edges):
...
+        
+        # Load star properties
+        ra, dec, e_psf, e_mod, de_psf, T_f, star_type = self.load_stars()
+
+        # Compute tau stats 
+        tau_stats  = {}
+        p_bestfits = {}
+
+        for t in STAR_TYPES:
+            if t==1:
```
joezuntz, line 224: With this break the loop is redundant, as only t=0 will be used.

txpipe/psf_diagnostics.py
```diff
@@ -172,6 +172,415 @@ def plot_histogram(self, function, output_name, xlabel, edges):
...
+                break
+            s = star_type == t
+    
+            # Load precomputed Rowe stats if they exist already
+            rowe_stats = self.load_rowe(t)
+
+            # Joint tau 0-2-5 data vector and cov
+            # tau_stats contains [ang, tau0, tau2, tau5, cov]
+            tau_stats[t] = self.compute_alltau(gal_ra, gal_dec, gal_g, gal_weight, s, ra, dec, e_psf, e_mod, de_psf, T_f)
```
joezuntz, line 233: can you use consistent underscoring in alltau and taustat - compute_all_tau and save_tau_stats.

txpipe/psf_diagnostics.py
```diff
@@ -172,6 +172,415 @@ def plot_histogram(self, function, output_name, xlabel, edges):
...
+                       "nbins"         : 20,
+                       "bin_slop"      : 0.01,
+                       "sep_units"     : "arcmin",
+                       "psf_size_units": "sigma",
+                     }
+
+    def run(self):
+        import treecorr
+        import h5py
+        import matplotlib
```
joezuntz, line 208: can you add `import emcee` up here - I try to do this so that all the modules that get imported happen at the start, so if there is anything missing you find out quickly.  Also then the version numbers are saved in any files output.

txpipe/psf_diagnostics.py
```diff
@@ -172,6 +172,415 @@ def plot_histogram(self, function, output_name, xlabel, edges):
...
+            rowe_stats[i] = theta, xi_plus, xi_err
+
+        return rowe_stats
+
+
+    def sample(self, tau_stats, rowe_stats, ranges, nwalkers=32, ndim=3):
+        '''
+        Run a simple mcmc chain to detemine the best-fit values for  
+        '''
+        import emcee,sys
```
joezuntz, line 271: It looks like sys isn't used.

txpipe/psf_diagnostics.py
```diff
@@ -172,6 +172,415 @@ def plot_histogram(self, function, output_name, xlabel, edges):
...
+        from scipy.stats import qmc
+
+        sampler = qmc.LatinHypercube(d=3, optimization="random-cd")
+        sample  = sampler.random(n=nwalkers)
+        qmc.discrepancy(sample)
```
joezuntz, line 276: This isn't printed or anything.

txpipe/psf_diagnostics.py
```diff
@@ -172,6 +172,415 @@ def plot_histogram(self, function, output_name, xlabel, edges):
...
+        corr0.process(catg, catp)
+        corr2 = treecorr.GGCorrelation(self.config)
+        corr2.process(catg, catq)
+        corr5 = treecorr.GGCorrelation(self.config)
+        corr5.process(catg, catw)
+        
+        # Estimate covariance using bootstrap ordering is xip,xim,xip,xim,xip,xim
+        # For our particular purpose we only care about xip so remove xim elements. 
+        cov = treecorr.estimate_multi_cov([corr0,corr2,corr5], 'bootstrap')
+        cov = np.delete(cov,np.arange(100,120),axis=0)
```
joezuntz, line 402: These numbers won't be correct if the user changes the nbins option.

txpipe/psf_diagnostics.py
```diff
@@ -172,6 +172,415 @@ def plot_histogram(self, function, output_name, xlabel, edges):
...
+        
+
+    def save_taustats(self, tau_stats, p_bestfits):
+        '''
+        tau_stats: (dict) dictionary containing theta,tau0,tau2,tau5 and cov
+        '''
+        f = self.open_output("tau_stats")
+        g = f.create_group("tau_statistics")
+        for s in STAR_TYPES:
+            if s==1:
```
joezuntz, line 424: Same note as above.

joezuntz (changes requested):

txpipe/psf_diagnostics.py
yomori: Fixed. Also moved `from scipy.stats import qmc`  and `import treecorr` here as well.

txpipe/psf_diagnostics.py
yomori: Removed.

txpipe/psf_diagnostics.py
yomori: Removed.

txpipe/psf_diagnostics.py
yomori: Fixed.

txpipe/psf_diagnostics.py
yomori: Actually, I tried moving this these imports to just below `run()` but it doesn't seem to propagate to the other functions so I get `'qmc' is not defined` and  `'treecorr' is not defined`

joezuntz: Yes, I usually have to put it in both of them. It doesn't slow anything down.

txpipe/psf_diagnostics.py
yomori: fixed

txpipe/psf_diagnostics.py
yomori: Ahh ok, I had thought you meant *move it* to just beneath run(), didn't realize you actually meant to leave it in the function as well. Fixed now.

txpipe/psf_diagnostics.py
yomori: Fixed.

txpipe/psf_diagnostics.py
yomori: Fixed.

yomori:
I remembered that there was an issue with the function `load_star_type()` so I fixed that as well, which was affecting how the stars were being loaded.  I also modified the code such that we only compute for a specific type of star (which now gets specified in the config using `star_type` = 'PSF-used' or 'PSF-reserved'). The reason for this is because for certain surveys they only provide the catalog for one type of star so it will automatically crash when it tries to compute tau for the other type.

joezuntz (approved):

PR #325, opened 2023-10-18
Stage to compute CL shear individual profiles and stacked profile, using CLMM functionalities

combet:


marina-ricci:
I started a new branch to test your new pipeline stage because i was lost in our different files. First thing I notice while trying to run is that you have `CLClusterShearCatalogs` defined in two files (`sources_select_compute.py` and `select.py`). That can't work. Is one file outdated?

marina-ricci:
I added some line of code and managed to run everything "pipeline wise" in my branch. I still need to check if the output makes sense. This is were I am :
image

marina-ricci:
If you look at `Run_CL_pipeline.ipynb`, you'll see the results of ensemble profile. I saved it to a file surely not in the smartest way but it works. Results make no sense (negative value), so I let you investigate that. Do you compute the ensemble profiles for all clusters in the cluster catalog? 
image

marina-ricci:
@combet : if I did not make a mistake, my new commits should allow you to run in pipeline mode. So you can pull from main and use the same template as for the binning notebook.

PR #331, opened 2023-11-03
Lognormal

elvinpoole:
Adds simulated lognormal fields from GLASS that can be contaminated with external weight map


joezuntz (changes requested):

txpipe/simulation.py
```diff
@@ -0,0 +1,379 @@
...
+                    has_rsd=False, 
+                    bias=(zb_grid,bz), 
+                    mag_bias=None
+                    )
+                )
+
+            self.ell = np.arange(self.lmax)
+            self.cls = []
+            for i in range(1,self.nshells+1):
+                for j in range(i, 0, -1):
```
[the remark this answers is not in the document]
elvinpoole: yes I agree, but I copied it from the glass code directly so i think it's right. I'm currently testing if the auto and cross correlation of the output bins matches the input theory, so if this is wrong it should show up there

txpipe/simulation.py
```diff
@@ -0,0 +1,379 @@
...
+                    cl_bin = cosmo.angular_cl(density[i-1], density[j-1], self.ell)
+                    self.cls.append( cl_bin )
+
+            #save the C(l)
+            cl_output = self.open_output("glass_cl", parallel=True)
```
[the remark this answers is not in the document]
elvinpoole: i removed the parallel=True here. Should i do the same for the other outputs e.g. photometry_catalog

txpipe/simulation.py
```diff
@@ -0,0 +1,379 @@
...
+        This method simulates galaxy positions within each matter shell and
+        populates them with corresponding redshifts and bin information
+
+        """
+        import glass.fields
+        import glass.points
+        import glass.galaxies
+        import healpy as hp 
+
+        if self.config["random_seed"] == "def":
```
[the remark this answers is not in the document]
elvinpoole: changed default seed to 0

txpipe/simulation.py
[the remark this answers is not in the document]
joezuntz: Yes, probably do that for now. If we parallelize this later we can update.

joezuntz:
You can just omit the "parallel" keyword altigether, and leave it as the default.

joezuntz:
Hi Jack - I forget, where are we on this one - is it ready?

elvinpoole:
not quite. Still figuring out why the output w(theta) doesn't match the input

elvinpoole:
now tested clustering is unbiased on nside=4096 (~0.8 arcmin) density maps

txpipe/simulation.py
```diff
@@ -0,0 +1,544 @@
...
+        chi_grid = np.arange(chi_min, chi_max + self.config["dx"], self.config["dx"])
+        a_grid = cosmo.scale_factor_of_chi(chi_grid)
+        self.zb = (1.0 / a_grid) - 1.0
+        self.ws = glass.shells.tophat_windows(self.zb, weight=camb_tophat_weight)
+        self.nshells = len(self.ws)
+        # TO DO: figure out why GLASS needed the linear ramp weight here
+
+    def get_bias_at_z(self, z):
+        bias0 = self.config["bias0"]
+        alpha_bz = self.config["alpha_bz"]
```
empEvil, line 123: Where is `bias0`, `alpha_bz`, and `zpivot` defined for the configs?

txpipe/simulation.py
```diff
@@ -0,0 +1,544 @@
...
+    inputs = [
+        ("mask", MapsFile),
+        ("lens_photoz_stack", QPNOfZFile),
+        ("fiducial_cosmology", FiducialCosmology),
+        ("input_lss_weight_maps", MapsFile),
+    ]
+
+    outputs = [
+        ("photometry_catalog", HDFFile),
+        ("lens_tomography_catalog_unweighted", TomographyCatalog),
```
empEvil, line 42: might be me that is dumb, but is this overwriting the photometry catalog or is it just adding to it/ making a new one?

empEvil (commented):
I have really only one concern, and that is I couldn't find where configs got defined, but other than that it looks good.

txpipe/simulation.py
elvinpoole: I added an example in examples/lognormal to show how the configs should be designed

empEvil: okay great! my main worry is that there aren't default values given, so what happens if none are provided? will the stage crash?

txpipe/simulation.py
elvinpoole: its making a new one. The idea is, this stage would be run at the start of the pipeline and its output gets used as the "photometry_catalog" (or just the ra,dec,z columns for now) for any later stages

txpipe/simulation.py
elvinpoole: yes it will return an error. I was reluctant to put in defaults because it will be very sample dependent that so someone using the stage probably *should* have to think about. happy to just put b0=2 as default though if you think thats better

empEvil: I think it might be better, and then add a big comment that people really should be adjusting these! that way people can at least go to the config list and see all the configs defined. But if you and Joe prefer it otherwise then leave it as is

elvinpoole: sounds good. I added the defaults

PR #333, opened 2023-12-05
mean shear vs mag

jmjefferson:
adding new diagnostic plot that looks at mean galaxy shear vs band magnitude


joezuntz:
This looks good. The error is unrelated. Merging.

PR #335, opened 2023-12-15
TXDensityMaps modifications

tmcornish:
1. Changed how delta_g is calculated.
2. Added option to exclude from the delta_g calculation any pixels whose values are below a user-specified threshold ("weight_thresh" in the config) in the mask .

### DIFF
--- a/txpipe/maps.py
+++ b/txpipe/maps.py
@@ -408,7 +408,7 @@ class TXDensityMaps(PipelineStage):
     """
     Convert galaxy count maps to overdensity delta maps
 
-    delta = (ngal - <ngal>) / <ngal>
+    delta = ngal / (weight * <ngal>/<weight>) - 1
 
     This has to be separate from the lens mappers above
     because it requires the mask, which is created elsewhere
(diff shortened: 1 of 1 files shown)


txpipe/maps.py
```diff
@@ -436,7 +439,9 @@ def run(self):
         mask[mask == healpy.UNSEEN] = 0
         mask[np.isnan(mask)] = 0
         mask = mask.flatten()
-        pix = np.where(mask > 0)[0]
+        #identify pixels to keep (i.e. mask value above specified threshold)
+        pix_keep = mask > self.config['weight_thresh']
+        pix = np.where(pix_keep)[0]
```
damonge, line 444: do we use `pix` for anything?

damonge (commented):
Just one comment from me!

txpipe/maps.py
tmcornish: It gets included in the output file but isn't used for any of the calculations in TXDensityMaps.

joezuntz:
This looks great - ignore the test fail, that's a JAX thing, not your problem.

Do you want to convert it from draft to a full PR?

joezuntz (approved):

PR #337, opened 2024-01-19
Ext cross corr

elvinpoole:
Add stage to compute two-point between external healsparse maps (e.g. survey properties or dust maps) and TXPipe tracers: Galaxy density and Shear

### DIFF
--- a/txpipe/twopoint.py
+++ b/txpipe/twopoint.py
@@ -253,8 +253,8 @@ class TXTwoPoint(PipelineStage):
         GAMMAT = sacc.standard_types.galaxy_shearDensity_xi_t
         GAMMAX = sacc.standard_types.galaxy_shearDensity_xi_x
         WTHETA = sacc.standard_types.galaxy_density_xi
-        POS_EXT = "pos_ext"
-        SHEAR_EXT = "shear_ext"
+        POS_EXT = sacc.build_data_type_name( "galaxy", ["density", "ext"], "xi" )
+        SHEAR_EXT = sacc.build_data_type_name( "galaxy", ["shear", "ext"], "xi" )
 
         comb = []
         for index, d in enumerate(results):
@@ -506,15 +506,11 @@ class TXTwoPoint(PipelineStage):
         elif k == POS_EXT:
             assert self.name == "TXTwoPointPixelExtCross"
             xx = self.calculate_pos_ext(i, j)
-            # TODO: make a new data type in sacc for external cross correlations
-            # for now we will use a string as a place holder
-            xtype = "pos_ext"
+            xtype = sacc.build_data_type_name( "galaxy", ["density", "ext"], "xi" )
         elif k == SHEAR_EXT:
             assert self.name == "TXTwoPointPixelExtCross"
             xx = self.calculate_shear_ext(i, j)
-            # TODO: make a new data type in sacc for external cross correlations
-            # for now we will use a string as a place holder
-            xtype = "shear_ext"
+            xtype = sacc.build_data_type_name( "galaxy", ["shear", "ext"], "xi" )
         else:
             raise ValueError(f"Unknown correlation function {k}")
 
--- a/txpipe/twopoint.py
+++ b/txpipe/twopoint.py
@@ -260,7 +260,7 @@ class TXTwoPoint(PipelineStage):
         for index, d in enumerate(results):
             # First the tracers and generic tags
             tracer1 = f"source_{d.i}" if d.corr_type in [XI, GAMMAT, SHEAR_EXT] else f"lens_{d.i}"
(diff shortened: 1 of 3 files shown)


txpipe/twopoint.py
```diff
@@ -250,12 +253,14 @@ def add_data_points(self, S, results):
         GAMMAT = sacc.standard_types.galaxy_shearDensity_xi_t
         GAMMAX = sacc.standard_types.galaxy_shearDensity_xi_x
         WTHETA = sacc.standard_types.galaxy_density_xi
+        POS_EXT = "pos_ext"
```
empEvil, line 256: Why not buiild a sacc data type for it? you can do something like `GAMMATS = sacc.build_data_type_name(
            "galaxy", ["shear", "Density"], "xi", subtype="ts"
        )` (Example pulled from the self-calibration extension

txpipe/twopoint.py
```diff
@@ -250,12 +253,14 @@ def add_data_points(self, S, results):
...
+        SHEAR_EXT = "shear_ext"
 
         comb = []
         for index, d in enumerate(results):
             # First the tracers and generic tags
-            tracer1 = f"source_{d.i}" if d.corr_type in [XI, GAMMAT] else f"lens_{d.i}"
-            tracer2 = f"source_{d.j}" if d.corr_type in [XI] else f"lens_{d.j}"
+            tracer1 = f"source_{d.i}" if d.corr_type in [XI, GAMMAT, SHEAR_EXT] else f"lens_{d.i}"
+            tracer2 = f"source_{d.j}" if d.corr_type in [XI] else (f"external_{d.j}" if d.corr_type in [POS_EXT, SHEAR_EXT] else f"lens_{d.j}")
```
empEvil, line 263: does this way of writing an `elif` actually work?

txpipe/twopoint.py
```diff
@@ -283,6 +288,7 @@ def add_data_points(self, S, results):
                         error=xiperr[i],
                         npair=npair[i],
                         weight=weight[i],
+                        tracers_later=tracers_later,
```
empEvil, line 291: what does this thing do? (if answered later disregard my message)

txpipe/twopoint.py
```diff
@@ -493,6 +503,18 @@ def call_treecorr(self, i, j, k):
         elif k == POS_POS:
             xx = self.calculate_pos_pos(i, j)
             xtype = sacc.standard_types.galaxy_density_xi
+        elif k == POS_EXT:
```
empEvil, line 506: Does this not cause issues if someone tries to run just TXTwopoint on an external, but not the class? i.e. Maybe we ought to move the entire function `call_treecorr` to the extension class

txpipe/twopoint.py
```diff
@@ -1092,6 +1114,235 @@ def calculate_pos_pos(self, i, j):
...
+            w=mask[pix], #weight pixels by their fracdet
+            k=map_d,
+            ra_units="degree",
+            dec_units="degree",
+            patch_centers=self.get_input("patch_centers"),
+        )
+        return cat
+
+
+    def write_output(self, source_list, lens_list, meta, results):
```
empEvil, line 1291: I think you need to provide something more to SACC to get it to actually save your file no?

empEvil (commented):
Overall looks good, but I am still trying to understand how you can get away with not defining the correlation type with SACC

txpipe/twopoint.py
elvinpoole: One of the tracers is the external map. This isn't a known tracer in sacc so it gives an error unless you set tracers_later=True

txpipe/twopoint.py
elvinpoole: seems to be saving for me. I'll add the example to demonstrate

txpipe/twopoint.py
elvinpoole: I didn't know you could do this! updated it

txpipe/twopoint.py
elvinpoole: i think it worked, but probably wasn't very clear. I made it a longer set of if and elif statements

txpipe/twopoint.py
elvinpoole: example added

txpipe/twopoint.py
empEvil: What do you think, should we add it to the subclass, or do you want to keep it here? might be worth to get @joezuntz  input once he is back?

empEvil (commented):
Think we just need to decide on what should be in the subclass and the main class. Other than that I think we are good

txpipe/twopoint.py
elvinpoole: ok sounds good. I moved it

elvinpoole:
removed all the edits to the base class and moved them to the new external cross correlation class

PR #340, opened 2024-02-17
PureB

yomori:
Implementation of hypbrideb and NaMaster in pureB to compute B-modes.


damonge:
Has anyone tested the standard B-mode purification method for cosmic shear data?
My expectation was always that it would not work at all, since it requires a smooth differentiable mask, which is very difficult to achieve for cosmic shear without losing most of the area.

yomori:
My gut feeling is the same as yours (I think the apodization scale has to be pretty wide too), which is why I didn't even bother. Might be able to do measure something if we apply constrained inpainting or something first (at which point the measurement isn't too meaningful).

txpipe/twopoint_null_tests.py
```diff
@@ -766,3 +767,272 @@ def write_output(self, source_list, lens_list, meta, results):
...
+            tmp       = np.array([])
+            ret=np.zeros((nbin_source,nbin_source,len(b.get_effective_ells())))
+
+            for zi in range(nbin_source):
+                for zj in range(zi,nbin_source):
+                     
+                    field1 = fields[zi]
+                    field2 = fields[zj]
+                    w_yp  = nmt.NmtWorkspace()
+                    w_yp.compute_coupling_matrix(field1, field2, b)
```
joezuntz, line 856: I think it's also useful to calculate the non-tomographic sample B mode.

txpipe/twopoint_null_tests.py
```diff
@@ -766,3 +767,272 @@ def write_output(self, source_list, lens_list, meta, results):
...
+    
+                    cl_coupled   = nmt.compute_coupled_cell(field1,field2)
+                    cl_decoupled = w_yp.decouple_cell(cl_coupled)
+                    ret[zj,zi,:]= cl_decoupled[3]
+                    tmp = np.concatenate([tmp,cl_decoupled[3]])
+
+            
+            # Compute covariance in a somewhat adhoc way by shuffling the pixel values.
```
joezuntz, line 864: Does this definitely work? Could also do the think where you randomly rotate shears.

txpipe/twopoint_null_tests.py
```diff
@@ -766,3 +767,272 @@ def write_output(self, source_list, lens_list, meta, results):
...
+            for k in tqdm(range(self.config['Nsims'])):
+                fields = {}
+                for i in range(nbin_source):
+                    idx1 = np.where(g1_maps[i]!=0)[0]
+                    idx2 = np.random.permutation(idx1)
+                    g1_maps[i][idx1]=g1_maps[i][idx2]
+                    g2_maps[i][idx1]=g2_maps[i][idx2]
+                    fields[i] = nmt.NmtField(mask, [g1_maps[i], g2_maps[i]], purify_e=False, purify_b=purify_b)
+
+                tmp = np.array([])
```
joezuntz, line 876: It doesn't really matter here, as all the arrays are small, but in general repeatedly concatenating arrays is less efficient than building up a list of arrays and concatenating them once.

But even more efficient (half memory usage) is making the full array at the start, which you can do when you know the final size as here.

Not a big deal here but it matters when the arrays are large.

txpipe/twopoint_null_tests.py
```diff
@@ -766,3 +767,272 @@ def write_output(self, source_list, lens_list, meta, results):
...
+                    for zj in range(zi,nbin_source):
+                        
+                        field1 = fields[zi]
+                        field2 = fields[zj]
+                        w_yp  = nmt.NmtWorkspace()
+                        w_yp.compute_coupling_matrix(field1, field2, b)
+        
+                        cl_coupled   = nmt.compute_coupled_cell(field1,field2)
+                        cl_decoupled = w_yp.decouple_cell(cl_coupled)
+                        tmp = np.concatenate([tmp,cl_decoupled[3]])
```
joezuntz, line 887: Not quite clear what this tmp array is for. It's only used to get the size in a moment, but then you can just accumulate that.

txpipe/twopoint_null_tests.py
```diff
@@ -766,3 +767,272 @@ def write_output(self, source_list, lens_list, meta, results):
...
+
+                covarr[:,k] = tmp
```
joezuntz, line 889: Confusing naming here - I was trying to work out why the covariance came directly from the C_ell. Should rename to make clearer.

txpipe/twopoint_null_tests.py
```diff
@@ -766,3 +767,272 @@ def write_output(self, source_list, lens_list, meta, results):
...
+             
+                    
+            self.ell     = b.get_effective_ells()
```
joezuntz, line 892: Do we need to store the windows as well as just the effective ell?

txpipe/twopoint_null_tests.py
```diff
@@ -766,3 +767,272 @@ def write_output(self, source_list, lens_list, meta, results):
...
+                geb_dict = {}
+                for i in range(0,self.config['Nell']):
+                    geb_dict['%d'%(i+1)] = {}
+                    for j in range(0,6):
+                        geb_dict['%d'%(i+1)]['%d'%(j+1)]=geb(i)[j]
+            
+                with open(file_precomputed_weights, 'wb') as handle:
+                    pickle.dump(geb_dict, handle, protocol=pickle.HIGHEST_PROTOCOL)
+
+            En0  = np.zeros(self.config["Nell"])
```
joezuntz, line 929: If you re-use a looked up variable many times like this it's clearer to pull it out once as `N_ell = self.config['N_ell']`, to reduce the line noise.

txpipe/twopoint_null_tests.py
```diff
@@ -766,3 +767,272 @@ def write_output(self, source_list, lens_list, meta, results):
...
+            # to precompute and load them in subsequent runs.
+            file_precomputed_weights = self.get_output("twopoint_data_fourier_pureB")+ f".precomputedweights.pkl"
+
+            if os.path.exists(file_precomputed_weights):
+                print(f"{self.rank} WARNING: Using precomputed weights from revious run: {file_precomputed_weights}")
+                with open(file_precomputed_weights, "rb") as f:
+                    geb_dict = pickle.load(f)
+            
+            else:
+                heb = hybrideb.HybridEB(self.config['theta_min'], self.config['theta_max'], self.config['Ntheta'])
```
joezuntz, line 917: Can you add more comments throughout this section? The main bit where it does the loop through pairs is okay but some more explanation of the various things we need first would be useful.

txpipe/twopoint_null_tests.py
```diff
@@ -766,3 +767,272 @@ def write_output(self, source_list, lens_list, meta, results):
...
+            nbin_source = max(len(source_tracers), 1)
+
+            corvarr = np.zeros( (int(nbin_source*(nbin_source+1)/2*self.config["Nell"]), self.config["Nsims"])  )
+
+            ret=np.zeros((nbin_source,nbin_source,self.config["Nell"]))
+
+            c = 0
+            for zi in range(0,nbin_source):
+                for zj in range(zi,nbin_source):
+                    ii = self.config["Nell"]*c
```
joezuntz, line 954: Explain what these indices are.

txpipe/twopoint_null_tests.py
```diff
@@ -766,3 +767,272 @@ def write_output(self, source_list, lens_list, meta, results):
...
+
+                    #random draws based on mean dn covariance
+                    x = np.random.multivariate_normal(mean = tmp.mean, cov =tmp.covariance.covmat , size = self.config["Nsims"])
+                    Rxip = x[:,:int(self.config["Ntheta"])]
+                    Rxim = x[:,int(self.config["Ntheta"]):]
+
+                    # Compute the uncertainties from random draws          
+                    for n in range(self.config["Nsims"]):
+                        for i in range(int(self.config["Nell"])):
+                            res     = geb_dict['%d'%(i+1)]
```
joezuntz, line 975: can do f"{i+1}" these days to be slightly more readable

txpipe/twopoint_null_tests.py
```diff
@@ -766,3 +767,272 @@ def write_output(self, source_list, lens_list, meta, results):
...
+
+            c = 0
+            for zi in range(0,nbin_source):
+                for zj in range(zi,nbin_source):
+                    ii = self.config["Nell"]*c
+                    ff = self.config["Nell"]*(c+1)
+                    
+                    # Need all 4 block of the covariance
+                    filename = self.get_input("twopoint_data_real_raw")
+                    tmp      = sacc.Sacc.load_fits(filename)
```
joezuntz, line 959: There's no need to read the sacc file in every time, you can read it once outside the loop and then `tmp.clone()` in the loop.

txpipe/twopoint_null_tests.py
```diff
@@ -766,3 +767,272 @@ def write_output(self, source_list, lens_list, meta, results):
...
+                        ell[i]  = res['4'][np.argmax(res['5'])]
+
+                    ret[zj,zi,:]= Bn0[:]
+                    c+=1
+                    
+            self.ell     = ell
+            self.results = ret
+            self.cov     = np.cov(corvarr)
+        else:
+            sys.exit("The only method implemented right now are 'hybrideb' or 'namaster_pureB' or 'namaster_nopureB' ")
```
joezuntz, line 999: You haven't imported `sys`, and in any case it's better to raise a ValueError so that if someone is using this in e.g. a notebook or other interactive use it won't kill their work.

txpipe/twopoint_null_tests.py
```diff
@@ -766,3 +767,272 @@ def write_output(self, source_list, lens_list, meta, results):
...
+
+        print('Saving pureB Cls in sacc file')
+        
+        self.save_power_spectra(nbin_source)
+
+
+    def save_power_spectra(self, nbin_source):
```
joezuntz, line 1006: Probably slightly more explicit to pass in results, ell, and cov to this function rather than save them as attributes of self.

txpipe/twopoint_null_tests.py
```diff
@@ -766,3 +767,272 @@ def write_output(self, source_list, lens_list, meta, results):
...
+        import sacc
+        import datetime
+
+        S = sacc.Sacc()
+        S.metadata['nbin_source'] = nbin_source
```
joezuntz, line 1011: Later we need to put the full provenance metadata in here, but that's a more general thing.

joezuntz (changes requested):

txpipe/twopoint_null_tests.py
yomori: Yes I am aware of this. This was just a sloppy way of getting the length.
I've removed the relevant lines and replaced the line where I initialize the array with 
`covarr = np.zeros((int(nbin_source*(nbin_source+1)/2*len(b.get_effective_ells())),self.config['Nsims']))
`

txpipe/twopoint_null_tests.py
yomori: Decided to directly get the array size with the known number of ell bins and numbers of zbin combination so this is not needed anymore. Removed.

yomori: This `tmp` is slightly different from the comment above. We need to estimate the full covariance that includes all the redshift combinations, which means we must create a matrix with each column being the full concatenated data vector and the different columns representing different realizations (to which we  apply np.cov later)

txpipe/twopoint_null_tests.py
yomori: Changed to `tmparr`.

txpipe/twopoint_null_tests.py
yomori: Fixed.

txpipe/twopoint_null_tests.py
yomori: Fixed. Also did the same for `Nsims` and `Ntheta`

examples/desy1/pipeline_Bmodes.yml
```diff
@@ -0,0 +1,116 @@
...
+    - name: TXSourceSelectorMetacal     # select and split objects into source bins
+    - name: TXShearCalibration          # Calibrate and split the source sample tomographically
+    #- name: TXTruthLensCatalogSplitter
+    #- name: TXStarCatalogSplitter  # Split the star catalog into separate bins (psf/non-psf)    
+    #- name: TXPhotozPlot 
+    - name: TXSourceMaps
+      #nprocess: 6
+      #threads_per_process: 1
+      #nodes: 1
+    #- name: TXLensMaps
```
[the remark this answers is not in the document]
yomori: As mentioned on the call, I will probably write one full systest pipeline yaml and clean stuff up in that one, but for now I removed all the unused stages.

txpipe/twopoint_null_tests.py
```diff
@@ -766,3 +767,272 @@ def write_output(self, source_list, lens_list, meta, results):
 
         # Finally, save the output to Sacc file
         S.save_fits(self.get_output("aperture_mass_data"), overwrite=True)
+
+
+
+class TXpureB(PipelineStage):
```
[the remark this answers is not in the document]
yomori: Fixed.

txpipe/twopoint_null_tests.py
```diff
@@ -766,3 +767,272 @@ def write_output(self, source_list, lens_list, meta, results):
...
+    '''
+    Make shear B-mode measurements
```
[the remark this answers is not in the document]
yomori: Added:

>    
    Make shear B-mode measurements.
    This stage computes xip and xim in very narrow angular bins and transforms them into 
    Fourier B-modes using narrow window functions. This allows us to compute B-modes without
    being affected by (a) the mask (which is an issue if we use https://arxiv.org/abs/astro-ph/0511629,
    which is implemented in NaMaster, since masks for LSS is more complicated than CMB) and
    (b) noise bias (since the measurements we make here are real-space measurements).

txpipe/twopoint_null_tests.py
```diff
@@ -766,3 +767,272 @@ def write_output(self, source_list, lens_list, meta, results):
...
+        import healpy as hp
+        import sacc
+        import pyccl
+        import hybrideb
+        import datetime
+        import pickle
+        from tqdm import tqdm
+
+        if self.config["method"] == 'namaster_pureB' or self.config["method"] == 'namaster_nopureB':
+            '''B-mode method that is already implemented in NaMaster'''
```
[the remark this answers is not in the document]
yomori: Fixed.

txpipe/twopoint_null_tests.py
```diff
@@ -766,3 +767,272 @@ def write_output(self, source_list, lens_list, meta, results):
...
+        import pymaster as nmt
+        import healpy as hp
+        import sacc
+        import pyccl
+        import hybrideb
+        import datetime
+        import pickle
+        from tqdm import tqdm
+
+        if self.config["method"] == 'namaster_pureB' or self.config["method"] == 'namaster_nopureB':
```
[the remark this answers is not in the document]
yomori: Actually the only difference between the two methods is the `purify_b` flag so if we try to separate the two methods, it would look identical except for the flag. Maybe I'm misunderstanding your suggestion.

txpipe/twopoint_null_tests.py
```diff
@@ -766,3 +767,272 @@ def write_output(self, source_list, lens_list, meta, results):
...
+            else:
+                print("WARNING: We are not running B-mode purification (this is what was used in DES-Y3).")
+                purify_b = False
+                
+            # Open source maps (g1,g2,weights)
+            with self.open_input("source_maps", wrapper=True) as f:
+                nbin_source = f.file["maps"].attrs["nbin_source"]
+                g1_maps     = [f.read_map(f"g1_{b}") for b in range(nbin_source)]
+                g2_maps     = [f.read_map(f"g2_{b}") for b in range(nbin_source)]
+                weights     = [f.read_map(f"lensing_weight_{b}") for b in range(nbin_source)]
```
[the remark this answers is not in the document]
yomori: I think this might be a scientific question, but I have removed this for now.

xuod:
@damonge @yomori I did try pure B-mode on DES Y3, but because of all the tiny holes in the mask, the apodization drastically reduces the effective area (or equivalently reduces the weight of many areas in the footprint). So that makes the test lose statistical power.

damonge:
Yep, indeed, this is the main problem with the standard purification approach. I could never convince myself that this would be useful for cosmic shear.

yomori:
Yeah so thats why I have three modes of running this right now.
(a) Namaster with `purify_b=False` -- to replicate DES-Y3 what did.
(b) Namaster with `purify_b=True` -- if anyone wants to run it on idealized sim or something (just a flag change anyway).
(c) HybridEB -- the more robust way  of computing BB.

If you are worried that someone will randomly run it with `purify_B=True`  without knowing the caveats and use that in a paper, I can throw in a warning. If you really really want it to be taken out, I'm happy to do that too.

txpipe/diagnostics.py
```diff
@@ -300,8 +300,11 @@ def plot_psf_shear(self):
         fig.close()
         
         f = self.open_output("g_psf_g_out")
-        data   =[mu1,mu2,mean11,mean12,mean21,mean22,std11,std12,std21,std22,line11,line12,line21,line22]
-        f.write(''.join([str(i) + '\n' for i in  data]))
+        data = np.c_[mu1,mu2,mean11,mean12,mean21,mean22,std11,std12,std21,std22,line11,line12,line21,line22]
```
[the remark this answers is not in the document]
yomori: Leaving as is for now since the use case for this is extremely limited. Happy to make amendments later switching to astropy.

txpipe/twopoint_null_tests.py
yomori: Fixed.

txpipe/twopoint_null_tests.py
yomori: Fixed.

txpipe/twopoint_null_tests.py
yomori: Fixed.

txpipe/twopoint_null_tests.py
yomori: Actually I'm finding that sacc.clone() doesn't exist, but .copy() does so I've switched to that.

txpipe/twopoint_null_tests.py
yomori: Added the following comments:

> 

            else:
            # To run this method we need to first compute the Fourier and real-space windows
            # hybrideb.HybridEB: Sets up the estimator by creating top hat filters in real-space
            # hybrideb.BinEB   : Calculates the intermediate filter quantities, namely fa, fb, M+, M- [Eqs 6-11]
            # hybrideb.GaussEB : Creates Gaussian windows in Fourier space
            heb = hybrideb.HybridEB(self.config['theta_min'], self.config['theta_max'], Ntheta)
            beb = hybrideb.BinEB(self.config['theta_min'], self.config['theta_max'], Ntheta)
            geb = hybrideb.GaussEB(beb, heb)
            
            # geb is in a fromat that can not be naturally saved so we convert it to a regular dictionary.
            geb_dict = {}
            for i in range(0,Nell):
                geb_dict[f"{i+1}"] = {}
                for j in range(0,6):
                    geb_dict[f"{i+1}"][f"{j+1}"]=geb(i)[j]
            
            # geb_dict stores the following information in a regular dictionary format
            # 1: theta in radians
            # 2: Fp  plus component of the real-space window function
            # 3: Fm  minus component of the real-space window function
            # 4: ell
            # 5: Wp  plus component of Fourier-space window function
            # 6: Wm  minus component of Fourier-space window function

            # 2 and 3 are used to get Xp/Xm      = sum((Fp*xip_hat +/- Fm*xim_hat)/2)    [eq1]
            # 5 and 6 are used to get <Xp>/<Xm>  = \int ell factors(ell) (Wp*Pe + Wm*Pb) [eq5]

            # Save this for subsequent runs since the windows don't change.

txpipe/twopoint_null_tests.py
yomori: Added the comment 

>              
    # Initial and final row indices of the resulting array (covarr) to fill.
    # For each tomographic bin we have Nell bins to fill.

txpipe/twopoint_null_tests.py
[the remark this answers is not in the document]
yomori: Since running NaMaster with `purify_b = False` is not really "PureB" I've rename this stage to TXShearBmode.

txpipe/twopoint_null_tests.py
yomori: I think what we typically do is rotate the shears at the catalog level not after projecting the galaxies onto pixels.
We would like to do that ideally but that requires passing down the full catalog to this stage and projecting iteratively. For now I've added pixel level rotation.

joezuntz: That's great, thanks!

PR #341, opened 2024-02-27
More configurable TXTwoPointFourier

tmcornish:
Users can now specify which components of 3x2pt analysis to run in TXTwoPointFourier. Computation of the theory power spectra has also been made optional. Certain inputs (fiducial_cosmology, source_noise_maps, lens_noise_maps, density_maps, source_maps) are now also optional depending on which analyses the user wants to run. 

Treatment of the survey mask has also been modified to account for cases where the mask is not binary. Users can now also specify a global config parameter, "mask_threshold", so that whenever the mask is read all pixels less than or equal to this value are automatically set to zero.

Tested with example 'metacal' and 'metadetect' pipelines, using all possible combinations of POS-POS, SHEAR-POS  and SHEAR-SHEAR.

### DIFF
--- a/examples/metacal/config.yml
+++ b/examples/metacal/config.yml
@@ -289,6 +289,10 @@ TXTwoPointFourier:
     ell_max: 150
     bandwidth: 20
     cache_dir: ./cache/workspaces
+    do_shear_shear: False
+    do_shear_pos: False
+    do_pos_pos: True
+    compute_theory: False
 
 TXSimpleMask:
     depth_cut : 23.5
--- a/examples/metacal/config.yml
+++ b/examples/metacal/config.yml
@@ -289,10 +289,10 @@ TXTwoPointFourier:
     ell_max: 150
     bandwidth: 20
     cache_dir: ./cache/workspaces
-    do_shear_shear: False
-    do_shear_pos: False
+    do_shear_shear: True
+    do_shear_pos: True
     do_pos_pos: True
-    compute_theory: False
+    compute_theory: True
 
 TXSimpleMask:
     depth_cut : 23.5
--- a/txpipe/twopoint_fourier.py
+++ b/txpipe/twopoint_fourier.py
@@ -90,6 +90,7 @@ class TXTwoPointFourier(PipelineStage):
         "do_shear_shear": True,
         "do_shear_pos": True,
         "do_pos_pos": True,
+        "compute_theory": True,
     }
 
     def run(self):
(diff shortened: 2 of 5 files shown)


examples/metacal/config.yml
elvinpoole, line 1: This changes the default behaviour of this example? do we want to do this? 
Could it be better to set them all to True? (I don't have a strong opinion either way)

txpipe/twopoint_fourier.py
```diff
@@ -116,28 +121,32 @@ def run(self):
...
-        # Note - this is currently unused, because we are using the noise
-        # maps instead of an analytic form, but that could change later
-        # so I will leave this here.
-        tomo_info = self.load_tomographic_quantities(nbin_source, nbin_lens, f_sky)
+        if not self.config["analytic_noise"]:
+            # Load in the per-bin auto-correlation noise levels and
+            # mean response values
+            # Note - this is currently unused, because we are using the noise
+            # maps instead of an analytic form, but that could change later
+            # so I will leave this here.
```
elvinpoole, line 148: Is this comment still correct? is the analytic form now implemented?

elvinpoole (commented):
Looks good to me. couple of very minor comments

txpipe/twopoint_fourier.py
elvinpoole: referring to this change in the config options
>     do_shear_shear: False
>     do_shear_pos: False
>     do_pos_pos: True
>     compute_theory: False

examples/metacal/config.yml
elvinpoole: referring to this change in the config options

>   do_shear_shear: False
>   do_shear_pos: False
>   do_pos_pos: True
>   compute_theory: False

tmcornish: You're right, I just forgot to reset them after testing - I'll set them all to True.

txpipe/twopoint_fourier.py
tmcornish: I think it's incorrect, and come to think of it it may actually have been for a while. Even before my changes, compute_noise_analytic didn't make use of the tomographic info loaded at this stage, so I wonder if it's left over from a while ago and is no longer relevant. Do you think it could be worth removing altogether?

elvinpoole: yes I think that makes sense

anicola:
Thanks a lot Tom. This looks good to me. Just a couple of questions/comments.

anicola:
Might it be safer to determine lmax based on the ell range specified to compute the spectra: https://github.com/LSSTDESC/TXPipe/blob/4885be178d975bcdeb762cad9d9b8a68d41b32d8/txpipe/twopoint_fourier.py#L128

anicola:
Might it be safer to determine lmax based on the ell range specified to compute the spectra instead of having hard-coded values? https://github.com/LSSTDESC/TXPipe/blob/4885be178d975bcdeb762cad9d9b8a68d41b32d8/txpipe/twopoint_fourier.py#L128

tmcornish:
Thanks for the comments @anicola. For some reason they're not showing up here but I saw them via email and have now replaced the hard-coded values in question with the user-provided ell_min and ell_max.

elvinpoole (approved):

anicola (approved):

PR #348, opened 2024-05-03
change macos version in CI

elvinpoole:
changing mac OS version in CI installation test

macos-latest now defaults to M1 which TXPipe does not yet officially support. Previously macos-latest pointed to macos-12 so fixing to that temporarily until we can get M1 installation set up


empEvil (approved):
We should, make a note in the readme that says we are doing this, so we will remember to undo it later, when NaMaster is fixed

PR #362, opened 2024-07-17
Save magnitudes in both source and lens catalogs

joezuntz:
The first step towards using RAIL on each tomographic bin separately is to save additional magnitude information to the per-bin lens and shear catalogs that we currently only use for 2pt measurement. This does that. The only fiddly part is getting all the names and prefixes correct - probably we should move some of that behaviour into the data type class.

As part of this I've also added metadata onto the various input catalogs listing what bands they include.


txpipe/calibrate.py
```diff
@@ -53,10 +54,21 @@ def run(self):
         extra_cols = [c for c in self.config["extra_cols"] if c]
         subtract_mean_shear = self.config["subtract_mean_shear"]
 
+        with self.open_input("shear_catalog", wrapper=True) as f:
+            bands = f.get_bands()
+
+        shear_prefix = self.config["shear_prefix"]
+        # this is the names of the columns in the input catalog
+        mag_cols_out = [f"mag_{b}" for b in bands] + [f"mag_err_{b}" for b in bands]
+        mag_cols_in = [f"{shear_prefix}{c}" for c in mag_cols_out]
```
elvinpoole, line 63: It looks like in the example the shear catalog input only contains 3 bands (the bands used in shear measurement maybe?) but for RAIL we will probably want all bands? 
Maybe either we should make sure the shear_catalog input has all available bands, or make the photometry_catalog magnitudes an input to this stage?

elvinpoole (changes requested):

txpipe/calibrate.py
joezuntz: It's still a bit up in the air whether there will be a full set of bands available in the shear catalog. For metadetect the mags have to be measured as part of the calibration process so that we can find the impact on the shear selection bias of cutting on them for tomography, so it's unfortunately not as simple as taking them from the photo catalog. So we might have to do shear tomography and n(z) estimation with a restricted set of bands.

txpipe/data_types/types.py
```diff
@@ -90,6 +101,16 @@ def get_primary_catalog_names(self, true_shear=False):
 
         return shear_cols, rename
 
+    def get_bands(self):
+        group = self.file[self.get_primary_catalog_group()]
+        if "bands" in group.attrs:
+            return group.attrs["bands"]
+        bands = []
+        for col in group.keys():
+            if col.startswith("mag_") and col.count("_") == 1:
```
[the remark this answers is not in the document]
joezuntz: That's a good point - thanks - I've assumed throughout TXPipe that we will have done that during the ingestion stage, but we don't want to re-run ingestion if we have multiple choices. So yes, we probably need a prefix. We'd need to update that throughout the entire codebase though, not just here. So if it's okay with you I'd like to do it all at once in a new PR.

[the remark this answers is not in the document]
elvinpoole: yes sounds good

txpipe/calibrate.py
elvinpoole: oh ok. I hadn't realized that. Then the current setup is good

elvinpoole (approved):

PR #374, opened 2024-10-18
Harmonize the column name for mag errors

joezuntz:


empEvil (commented):
Looks good, I see no issues with this. I am a little confused though, It only looks like we are loading in the `mag_err` in ingest part of the pipeline, but are we actually using it somewhere? if no, why are we loading it? if yes, does the name change need to be there as well?

empEvil:
that is a fair point, it is good for future proofing. I was just curious

PR #377, opened 2024-11-28
Cluster clustering measurement code

AlessandraFumagalli:
Wrote code for measuring 2D two-point correlation function of clusters, in richness bins. 
Wrote notebook to test the code and validate it against and external code.


examples/clustering_test/config/config_clusters.yml
joezuntz, line 1: Could you give the directory here ("clustering_test") a slightly more specific name?

examples/clustering_test/outputs/test/cluster_twopoint_real.sacc
joezuntz, line 1: If this is an output file then you also don't want to include it I think?

txpipe/__init__.py
```diff
@@ -39,6 +39,7 @@
 from .lssweights import TXLSSWeights
 from .simulation import TXLogNormalGlass
 from .magnification import TXSSIMagnification
+from .twopoint_cluster import TXTwoPointCluster
```
joezuntz, line 42: If we move the twopoint_cluster file (see comment below) then you should also change this.

txpipe/twopoint_cluster.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        Save the correlation function results to a SACC file.
+        """
+        s = sacc.Sacc()
+
+        for result in results:
+            meanr, xi, varxi, tracer1, tracer2 = result
+
+            # Add tracers (placeholders z/Nz values)
+            s.add_tracer('NZ', tracer1, np.array([1]), np.array([1]))
```
joezuntz, line 92: Is this just a placeholder? You can also use a "misc" tracer for this. I think each tracer here is a richness/redshift bin - is that right?  We might want to find a way to represent that.

joezuntz (changes requested):
Looks goof in general - see notes below for requested changes.

txpipe/__init__.py
AlessandraFumagalli: Not clear what is the "comment below".
Where should I move the twopoint_cluster file?

examples/twopoint_cluster/TXTwoPointCluster.out
```diff
@@ -0,0 +1,2 @@
+Executing stage: TXTwoPointCluster @ 2025-01-13 17:35:10.802489
```
joezuntz, line 1: Can you remove this file from the PR?

examples/twopoint_cluster/outputs/test/cluster_twopoint_real.sacc
joezuntz, line 1: This file should also be removed from the PR

joezuntz (commented):
I've mentioned the only two things I'd like you to do from a repo tidiness point of view.  Would it be useful for me to do a more thorough (code and scientific) review for you or would you just like to manage that internally within the CL group?  This is all in a subdirectory so I'm happy either way, as long as nothing breaks the main code.

txpipe/extensions/twopoint_cluster.py~
joezuntz, line 1: Sorry - I think I missed this one last time - it's an empty ~ file left over by an editor - can you delete?

joezuntz: Actually never mind, I can delete it.

txpipe/extensions/twopoint_cluster.py
```diff
@@ -0,0 +1,181 @@
+import sys
+import os
+import numpy as np
+import treecorr
+import h5py
+import sacc
+from txpipe.base_stage import PipelineStage
```
joezuntz: The right syntax for importing from a different sub-package is to use a relative import:

```
from ..base_stage import PipelineStage
from ..data_types import HDFFile, SACCFile
```

I have updated this.

joezuntz:
I've made some small tweaks to this and now can merge it once the tests are run.

joezuntz (approved):

PR #378, opened 2024-12-10
SSI depth maps

elvinpoole:
- Adds a stage `aux_ssi_maps` to create depth maps from SSI catalogs, using both true and measured magnitudes to estimate depth

- Modifies the map plotting script to plot the new SSI depth maps

- Reorganizes the SSI ingestion stage to have simpler class inheritance


txpipe/data_types/types.py
```diff
@@ -274,12 +274,20 @@ def write_map(self, map_name, pixel, value, metadata):
         subgroup.create_dataset("pixel", data=pixel)
         subgroup.create_dataset("value", data=value)
 
-    def plot_healpix(self, map_name, view="cart", **kwargs):
+    def plot_healpix(self, map_name, view="cart", rot180=False, **kwargs):
```
elvinpoole, line 277: Here I added an option to rotate the map by 180 degrees in the RA direction, which I found useful for plotting surveys that cross the RA=0 line

joezuntz (approved):

PR #380, opened 2025-01-20
Documentation Update

joezuntz:
This adds partial or full documentation, mainly on files and file types.

It also updates various docstrings to make the sphinx-friendly.

Finally, it updates the list of stages.


docs/src/file_details/binned.rst
```diff
@@ -0,0 +1,95 @@
...
+Binned shear catalogs
+---------------------
+
+The tomographically binned shear catalogs also include the shear calibration to correct for measurement and selection biases. 
+
+The bin index b runs from 0 to nbin-1, or "all" tag for a combined non-tomographic sample. The "all" bin is not simply the same as all the other bins combined, because the selection bias calibration is different.
+
+Like the original source shear catalogs, the binned shear catalogs may include only a subset of magnitude values depending on what was measured during the pipeline.
+
+=======  =======  =========  ==========  =========
```
empEvil, line 18: So this is an example of what a file could look like?

docs/src/file_details/metadata.rst
empEvil, line 1: Little barebones

docs/src/file_details/supreme.rst
empEvil, line 1: I know this if for future use, but shouldn't we say that then?

docs/src/nersc.rst
empEvil, line 1: I think it would be good to make it even clearer that the ``tx`` environment is deprecated

txpipe/calibrate.py
empEvil, line 1: Why remove details?

empEvil (commented):
I think we also need to update the file `running.rst` since it references running on nersc on `cori`.

docs/src/file_details/binned.rst
joezuntz: Yep - I've added a note to this effect.

txpipe/calibrate.py
joezuntz: Firstly because the layout of them with the bullet points was causing an error in the docs!  But looking at it it seemed like an old discussion that isn't really relevant now.

joezuntz:
Removed the mentions of "cori" too!

empEvil (approved):
Great!

joezuntz:
Thanks - I fixed that broken link.

PR #384, opened 2025-01-29
Remove FlexZPipe and references

joezuntz:
FlexZBoost is now included via RAIl so the old unused submodule can be removed. One pipeline still used it but was broken anyway so I have removed that too (redmagic "complete" pipeline - I don't even recall what it did differently to the main redmagic pipeline).


joezuntz:
Deleting config and submod only so not requesting review.

PR #388, opened 2025-03-04
SSI depth maps from detection fraction

elvinpoole:
adds SSI depth maps defined as the magnitude below which X% of objects are detected in each healpix pixel


joezuntz (changes requested):
Just some minor comments.

joezuntz (approved):

PR #390, opened 2025-04-22
Changes to run RAIL on CosmoDC2 at scale

joezuntz:
These are mostly configuration and installation changes, but also include:

- fixes to save the magnitudes correctly when saving tomographic shear catalogs
- a parallelization fix for calling RAIL multiple times


bin/perlmutter-install.sh
```diff
@@ -39,12 +21,12 @@ rm -rf ${FIRECROWN_DIR}
 
 # Add some things that are set up automatically
 cat >  ./conda/etc/conda/activate.d/activate-txpipe.sh <<EOF
-module load cray-mpich-abi
+module load mpich/4.2.2
```
empEvil, line 24: Is this getting loaded twice? line 24 and line 8?

examples/cosmodc2/config.yml
```diff
@@ -1,16 +1,7 @@
 global:
     pixelization: healpix
     chunk_rows: 1000000
```
empEvil, line 3: Is this not getting defined twice?

empEvil (commented):
Looks good, some minor comments. One thing, the changes to `types.py` and `calibrate.py` we are sure they work also if we run something that isn't metacal or meta detect?

examples/cosmodc2/config.yml
empEvil: never mind one is `chunk_rows` and other is `chunk_size`, why do we have both?

joezuntz: The RAIL people use `chunk_size` and TXPipe uses `chunk_rows`. At some point we should probably just update TXPipe to copy RAIL, there's fewer repos to change for us.

bin/perlmutter-install.sh
joezuntz: The one on line 24 is not being executed here, it's being written to `./conda/etc/conda/activate.d/activate-txpipe.sh`.  I've added a comment to explain, and made it clearer with indentation.

joezuntz:
The CI checks that lensfit and null calibrations both work. The behaviour shouldn't have changed here if `shear_prefix` is not set. Hopefully that's enough.

empEvil (approved):

PR #391, opened 2025-04-23
Add Filtered Square Bispectrum calculation

joezuntz:
Thanks to @ashaeres for adding!


joezuntz:
We wrote this extension code together so it's already been reviewed as it was written. So merging now.

PR #403, opened 2025-08-28
Use spectroscopic RAIL file instead of old calib table

joezuntz:
The selectors have long used an old legacy ascii table that replicated RAIL's spectroscopic files. This changes them to read from those HDF files instead.


elvinpoole (approved):
looks good

PR #411, opened 2025-09-24
added git to dataregistry pip install at nersc

elvinpoole:
I found the perlmutter install wasn't working because of a missing git+ in the install yml file. This PR just adds it


joezuntz (approved):
Thanks! I had experienced this earlier today but failed to fix it.

PR #414, opened 2025-10-02
Minor fixes to DP1 runs

elvinpoole:
- I found i needed the latest development sacc to get the two-point functions to save. I temporarily added the latest commit to the environment files until there is a new sacc version

- also removed the ggl null tests from pipeline-edfs.yml to match pipeline.yml


joezuntz:
Sorry about the sacc thing - I forgot to do the release. That's now out, and version 2.0.1 should work with this.

joezuntz:
Then we can just do the GGL change

PR #454, opened 2026-03-24
Firecrown 1.14

joezuntz:
This updates to being compatible with (and I assume) requiring FireCrown 1.14


elvinpoole (changes requested):
Firecrown edits look good, but the NERSC install has an conflict with the numcosmo_py

joezuntz:
We may need to put the numcosmo dependency in the conda package

PR #464, opened 2026-04-23
Remove the mcal_ prefix and standardize on the psf_ prefix

joezuntz:
The mcal_ prefix in some shear catalogs was inherited from the mcal input catalogs but ended up just confusing things. This removes it from wherever it's used in:
- ingestion when creating TXPipe files
- usage so you no longer specify the `shear_prefix` and `psf_prefix` options

A new version of the example data file has been created that removes the prefix from the shear catalogs, and the script I used to do this is also included.

We should re-generate other data files (e.g. at NERSC) or run this script on them. My next planned work is a rebuild of the ingestion so after that we should just re-run everything anyway (and put it in the data registry once ingested).


txpipe/diagnostics.py
```diff
@@ -85,7 +84,7 @@ def run(self):
                 # Create dask arrays of the columns. This loads them lazily,
                 # so no data is actually loaded here. Only when the "compute"
                 # method is called below does anything actually happen.
-                col = da.from_array(f[f"shear/{old_name}"], chunks=chunk_rows)
```
empEvil, line 88: This is not a shear prefix being removed, but rather just the shear, which I thought was a grouping in the HDF5 file? did we remove that grouping as well?

empEvil (commented):
Looks fine, except one line where I am not sure we are removing the right thing, or I misunderstood something

txpipe/diagnostics.py
joezuntz: Right, so, because we no longer ask the user for the shear prefix which would have been something like "mcal_" or "00/" we now use the `get_primary_catalog_group` method above on the `ShearCatalog` object. But that gives the full group path, including the "/shear".  So above in all the dictionaries mapping names we now use that group name, which means we can skip all the "shear/" parts here because they are already included.

empEvil (approved):

PR #465, opened 2026-04-28
Selection functions from ingested SSI catalogues

tmcornish:
Added two stages to `auxiliary_maps.py`:
- `TXSelectionFunctionSSIMaps`: Uses ingested SSI catalogues to estimate the survey selection function and its 1-sigma uncertainties in pixels containing injected sources. This makes use of a new function called `make_dask_selection_function` added to `dr1.py`.
- `TXModelSelectionFunction`: Uses a model to extrapolate the selection function to the rest of the footprint, as defined via an input mask. For now this model is a multilinear fit to survey property maps, but there is scope for alternative models to be added later.


tmcornish:
The current version is a basic working version and has been tested on DES Y3 SSI catalogues, but there are still a few TODOs to deal with before it's ready for merging:
- The selection function in a given pixel is currently just estimated as the ratio of detected injections to the total number of injections in that pixel. Incorporating some means of applying lens sample-like cuts will give more useful estimates for clustering analyses.
- Uncertainties on the selection function are approximated as Gaussian, but this breaks down if the number of detections in a pixel is either close to 0 or close to the total number of injections in that pixel. 
- Need to make `TXModelSelectionFunction` dask-compatible
- Currently the best-fit parameters of the model and their covariance matrix are not saved anywhere.

tmcornish:
Major changes since the last comment:
- Removed `TXSelectionFunctionSSIMaps` since it was essentially doing a subset of what TXAuxiliarySSIMaps` already did.
- Modified `TXAuxiliaryMaps` so that it also outputs the total injections per pixel as a function of magnitude threshold (this is so that it can be used in error estimation later on)
- Modified `TXModelSelectionFunction` to instead use outputs from `TXAuxiliaryMaps` as inputs, namely the `frac_det_by_mag_thresh` and the (newly added) `inj_count_by_mag_thresh` maps. This stage now creates model predictions for each magnitude threshold.
- `TXModelSelectionFunction` is now dask-compatible.
- The best-fit parameters of the model and their covariance matrices are now saved to an HDF5 file.
- Added config options for exclusion of pixels below a given coverage fraction and number of total injections. 

I also found a bug where any attempt to load a mask with a non-zero threshold for removing pixels would produce an empty map, which should now be fixed.

txpipe/auxiliary_maps.py
```diff
@@ -459,3 +461,330 @@ def run(self):
             for map_name, m in hsp_maps.items():
                 out.write_map(map_name, m, metadata)
             out.file["maps"].attrs.update(metadata)
+
+
+class TXModelSelectionFunction(TXBaseMaps):
```
joezuntz: Could you use a slightly more specific name than this - model selection is quite broad.

txpipe/auxiliary_maps.py
```diff
@@ -459,3 +461,330 @@ def run(self):
...
+
+        with self.open_input("mask", wrapper=True) as f:
+            mask = f.read_mask(
+                thresh=self.config["mask_thresh"],
+                degrade_nside=pixel_scheme.nside
+            )
+        mask_pix = mask.valid_pixels
+
+        # Load survey property maps at the valid pixels
+        # TODO: Define these as individual inputs rather than glob search?
```
joezuntz, line 554: We've used both approaches in the past. We should probably standardize on one.

txpipe/auxiliary_maps.py
```diff
@@ -459,3 +461,330 @@ def run(self):
...
+        samples = sel_func_meas.dtype.names
+        maps = {
+            "sel_func": [],
+            "err_sel_func": []
+        }
+        model_info = {
+            "alphas": [],
+            "cov_alphas": []
+        }
+        for k in samples:
```
joezuntz, line 591: You need a little error handling here to make the error message easier if there is no overlap or no pixels with enough injections.

txpipe/auxiliary_maps.py
```diff
@@ -459,3 +461,330 @@ def run(self):
...
+        pixel_scheme = choose_pixelization(**self.config)
+
+        # Load the map of the measured selection function and its uncertainties
+        with self.open_input("aux_ssi_maps", wrapper=True) as f:
+            sel_func_meas = f.read_map("depth_det_prob/det_frac_by_mag_thres")
+            ninj = f.read_map("depth_det_prob/inj_count_by_mag_thres")
+        # Ensure correct resolution (desired resolution can only be lower than
+        # or equal to the native resolution of the selection function map).
+        # "sum" reduction used for ninj since this is a counts map
+        sel_func_meas = sel_func_meas.degrade(pixel_scheme.nside)
```
elvinpoole: Should this degrade use a weighted mean reduction instead? (I think the weight would be injected number counts?)

txpipe/auxiliary_maps.py
```diff
@@ -459,3 +461,330 @@ def run(self):
...
+        print(f"Found {nsys} total systematic maps")
+        spmaps = [
+            hsp.HealSparseMap.read(
+                sf,
+                degrade_nside=pixel_scheme.nside
+            )
+            for sf in sysfiles 
+        ]
+        
+        # Identify valid pixels across mask and all SP maps
```
elvinpoole, line 567: For higher resolutions, using healsparse's own intersection tools might be faster here
something like:
```
goodmap = hsp.operations.product_intersection([mask, sel_func_meas]+spmaps)
goodpix = goodmap.valid_pixels
```
and the values of goodmap aren't ever used, just the pixels

txpipe/auxiliary_maps.py
```diff
@@ -459,3 +461,330 @@ def run(self):
...
+    ]
+
+    config_options = {
+        "block_size": StageParameter(int, 0, msg="Block size for dask processing (0 means auto)."),
+        "systmaps_dir": StageParameter(str, "", msg="Directory containing systematic maps."),
+        "degree": StageParameter(int, 1, msg="Degree of the polynomial fit."),
+        "mask_thresh": StageParameter(float, 0.0, msg="Threshold for masking pixels at native resolution of mask."),
+        "mask_thresh_coarse": StageParameter(float, 0.0, msg="Threshold for masking pixels after mask is degraded."),
+        "inj_count_thresh": StageParameter(int, 1, msg="Exclude pixels containing fewer injections than this number."),
+        "sel_func_err_type": StageParameter(str, "none", msg="Type of uncertainty attributed to measured selection function."),
```
elvinpoole: we should list the options of possible `sel_func_err_type` in the msg here

txpipe/auxiliary_maps.py
```diff
@@ -459,3 +461,330 @@ def run(self):
...
+        da = self.da
+        # Degree of polynomial fit
+        deg = self.config["degree"]
+
+        # Weights for each training pixel
+        W = self.sel_func_weights(yerr_train)
+        # If Gaussian approx was used for uncertainty estimation, pixels in
+        # which the selection function is 0 or 1 will have zero uncertainty
+        # (and thus infinite weight); remove these pixels here.
+        (keep,) = da.compute(da.isfinite(W))
```
elvinpoole, line 837: it might be possible to avoid this early da.compute here,  maybe with `da.where`? not sure if there would be much performance improvement from it though, so feel free to ignore if you think this is more optimal

txpipe/auxiliary_maps.py
```diff
@@ -459,3 +461,330 @@ def run(self):
...
+        cov_alphas = np.linalg.inv(XtWX)
+
+        # Now make predictions at X_pred
+        m = len(X_pred[:,0])
+        ones = da.ones((m, 1), chunks=(X_pred.chunks[0], 1))
+        X = da.hstack([ones, X_pred])
+        for n in range(2, deg + 1):
+            X = da.hstack([X, da.power(X_pred, n)])
+        y_pred = alphas @ X.T
+        err_y_pred = np.sqrt(np.diag(X @ cov_alphas @ X.T))
```
elvinpoole: what shape is X and how large could it get? If large, could this step be a bottleneck?

txpipe/auxiliary_maps.py
```diff
@@ -459,3 +461,330 @@ def run(self):
...
+        else:
+            raise ValueError(
+                "err_type must be either 'none' or 'gaussian'."
+            )
+
+    def sel_func_weights(self, err_sel_func):
+        """
+        Converts selection function uncertainties into weights for modelling.
+        """
+        if all(err_sel_func == 0):
```
elvinpoole: i think this is an implicit compute of the dask array (might not be an issue, just pointing it out in case it can be optimized)

txpipe/auxiliary_maps.py
elvinpoole, line 591: so this current runs on every map in depth_det_prob/det_frac_by_mag_thres. We might want to add something that lets you specify just one of those maps to run on?

elvinpoole (commented):
This looks great! 

It might be useful to also upload an example ini file (or just add this stage to the existing SSI depth one)
Also, some of the methods (particularly polynomial_model) could do with a bigger docstring so we can see the inputs are dask arrays and what their expected shape will be
In general I wonder if the current dask logic is optimal. There are a few places where  da.compute is called early so you lose the laziness of the dask arrays. This might be fine, just bringing it up in case there is a better way....

txpipe/auxiliary_maps.py
tmcornish: You're right, it should be weighted mean – fixed!

joezuntz (commented):
A couple of comments from me too.

txpipe/auxiliary_maps.py
tmcornish: Done! Although I've excluded `sel_func_meas` as `goodmap` should cover the whole footprint (as defined by the mask and survey property maps), whereas `sel_func_meas` is limited to regions where SSI has been done.

txpipe/auxiliary_maps.py
tmcornish: Done!

txpipe/auxiliary_maps.py
tmcornish: I've added a config option allowing the user to list which bins to run on

txpipe/auxiliary_maps.py
tmcornish: This seems to run into issues when using `keep` to slice the dask arrays in the following lines (I get complaints about unknown array sizes), so I've kept it as is for now.

txpipe/auxiliary_maps.py
tmcornish: I've now replaced this with a cheaper equivalent which avoids computing the whole matrix multiplication

txpipe/auxiliary_maps.py
tmcornish: I've now circumvented this by using the value of `self.config["sel_func_err_type"]` as the condition instead

txpipe/auxiliary_maps.py
tmcornish: Changed to `TXPredictDensitySelectionFunction`

txpipe/auxiliary_maps.py
tmcornish: Added a line to assert that the number of valid pixels > 0, along with an error message for if this isn't the case

txpipe/auxiliary_maps.py
tmcornish: I've left it as is for now but will open an issue to note that it should be standardised at some point

elvinpoole (approved):
looks great! I'm happy with all the parts I previously commented on

joezuntz (approved):

PR #466, opened 2026-04-29
DES-Y3 3x2pt

joezuntz:
This is an attempt at pulling everything together for a full 3x2pt re-run of DES Y3.

It's been run on a sub-deg2 test patch and now I'm looking at pushing it through the full catalog on NERSC.

The main remaining TODO apart from the full run is to try to improve the photo-z treatment. RAIL team members have pointed me towards some runs on Cardinal. It would be nice to try to incorporate this once the current attempt to process the full catalog is done.

This would also be a chance to try and use the RAIL tomography steps.


txpipe/lssweights.py
```diff
@@ -1111,38 +1129,52 @@ def compute_weights(self, density_correlation):
...
+                map_index=sig_map_index,
+                frac=frac_map,
+                do_grid_hist=False
+            )
+            assert np.allclose(
+                Fdc.ndens,
+                density_correlation.linear_model(coeff),
+                rtol=1e-10,
+                atol=1e-12,), "linear model output ndens differs between design_matrix computation computing from the maps directly"
+
```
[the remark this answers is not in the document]
joezuntz: This only happens once per tomo bin so it's negligible compared to the time taken for the minimization.

txpipe/lsstools/lsstools.py
```diff
@@ -586,25 +693,28 @@ def linear_model(
     Fvals = np.sum(np.array([alphas]).T * sysmap_table[map_index], axis=0) + beta
     F.update_values_pix(validpixels, Fvals)
```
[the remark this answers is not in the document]
joezuntz: There are combinations that will fail here but they are not used.

examples/desy3/pipeline.yml
```diff
@@ -88,22 +172,16 @@ config: examples/desy3/config.yml
...
     fiducial_cosmology: data/fiducial_cosmology.yml
-    random_cats_source: Null
-    shear_tomography_classifier: none
-    flow: data/example/inputs/example_flow.pkl
+    desy3_index_catalog: data/desy3/inputs/DESY3_indexcat.h5
 
 # if supported by the launcher, restart the pipeline where it left off
 # if interrupted
-resume: False
+resume: resume
```
empEvil, line 184: Just double checking, it can be either resume or True?

txpipe/ingest/stage3/desy3.py
```diff
@@ -146,3 +160,158 @@ def run(self):
...
+    outputs = [
+        ("shear_catalog", ShearCatalog),
+    ]
+
+    config_options = {
+        "input_group_name": StageParameter(str, "catalog/", msg="Input group name in the HDF5 file."),
+        "chunk_rows": StageParameter(int, 100_000, msg="Number of rows to process in each chunk."),
+    }
+
+    def des_metacal_flux_to_mag(self, flux, flux_err):
```
empEvil, line 181: How is this different, than what we do with flux to mag in utils?

txpipe/source_selection/metacal.py
```diff
@@ -61,7 +79,6 @@ def setup_output(self):
             R_gamma: the per-object estimator response
             R_S: the per-bin selection response
             R_gamma_mean: the mean per-bin estimator response
-            R_total: the complete per-bin response
```
empEvil, line 64: is there a reason it is not getting saved any longer?

empEvil (commented):
Looks good

examples/desy3/pipeline.yml
joezuntz: It can be either!
- `resume` = start where you left off
- `restart` = start from the beginning
- `refuse` = don't re-start if any files already exist
- `True` = alias for `resume` for backwards compatibility
- `False` = legacy for `restart` for backwards compatibility

I need to put together more ceci docs!

txpipe/source_selection/metacal.py
joezuntz: basically because it was redundant and never used. It is just equal to R_gamma + R_S, and in the calibrator we were just loading the latter two.

txpipe/ingest/stage3/desy3.py
joezuntz: They have different reference fluxes / magnitude zero-points. So there is a constant offset between the magnitudes or equivalently a constant ratio between the fluxes.

Either is allowed as long as the things you compare to are done consistently.

joezuntz: I'll add a comment here!

joezuntz:
@empEvil think I've fixed all this - can you approve?

empEvil (approved):

PR #469, opened 2026-05-05
MetaDetect Ingester for DP2

empEvil:
This is the ingestion stage for the Rubin MetaDetect catalog as we expect it to look for DP2


empEvil:
It currently assumes 'ns' rather than '00' so we might need to merge that change at the same time

empEvil:
so in the [Schema](https://sdm-schemas.lsst.io/lsstcam.html#ShearObject) for MetaDetect there are 2 different `psf_T` values, one is the original, and one is after the reconvolution, which one is it we require?

joezuntz:
Good question. I guess for computing T / Tpsf we need the re-convolved one since the T will be re-convolved as well? But for null tests we probably want the other one. I guess this is why both were in there before. That's annoying. Could you ingest the non-reconvolved one as "original" or something like that?

txpipe/ingest/MetaDetect.py
```diff
@@ -0,0 +1,324 @@
...
+    "deepCoadd_sky_background_consolidated_map_weighted_mean": "Weighted mean of background light level from the sky (nJy)",
+    "deepCoadd_sky_noise_consolidated_map_weighted_mean": "Weighted mean of standard deviation of the sky level (nJy)",
+    "deepCoadd_dcr_dra_consolidated_map_weighted_mean": "Weighted mean of DCR-induced astrometric shift in right ascension direction, expressed as a proportionality factor",
+    "deepCoadd_dcr_ddec_consolidated_map_weighted_mean": "Weighted mean of DCR-induced astrometric shift in declination direction, expressed as a proportionality factor",
+    "deepCoadd_dcr_e1_consolidated_map_weighted_mean": "Weighted mean of DCR-induced change in PSF ellipticity (e1), expressed as a proportionality factor",
+    "deepCoadd_dcr_e2_consolidated_map_weighted_mean": "Weighted mean of DCR-induced change in PSF ellipticity (e2), expressed as a proportionality factor",
+}
+
+
+class TXIngestMetaDetect(PipelineStage):
```
joezuntz: Can you add DP1 to the name of this class to distinguish from e.g. DESY6.

Also since all of this recaps some stuff that is in dp1.py, is there a reason it's not in there, or deleted from there?

txpipe/ingest/__init__.py
```diff
@@ -11,3 +11,4 @@
 )
 from .dp1 import TXIngestDataPreview1
 from .stage3 import TXIngestDESY3Gold
+from .MetaDetect import TXIngestMetaDetect
```
joezuntz: file names should generally start with lower case.

txpipe/ingest/lsst.py
```diff
@@ -64,3 +64,55 @@ def process_shear_data(data):
...
+    output = {}
+    for variant in ["ns", "1p", "1m", "2p", "2m"]:
+        var_data = data[data["metaStep"] == variant]
+        var_data = sanitize(var_data)
+
+        var_output = {
+            "ra": var_data["ra"],
+            "dec": var_data["dec"],
+            "id": var_data["shearObjectId"],
+            "metaStep": var_data["metaStep"].astype("S"), #might not be needed
```
joezuntz: Should all be the same I think?

txpipe/ingest/lsst.py
```diff
@@ -64,3 +64,55 @@ def process_shear_data(data):
...
+            "T": var_data["gauss_T"],
+            "s2n": var_data["gauss_snr"],
+            "T_err": var_data["gauss_TErr"],
+            "psf_g1_original": var_data["psfOriginal_e1"],
+            "psf_g2_original": var_data["psfOriginal_e2"],
+            "psf_T_mean_original": var_data["psfOriginal_T"],
+            "psf_g1": var_data["gauss_psfReconvolved_g1"],
+            "psf_g2": var_data["gauss_psfReconvolved_g1"],
+            "psf_T_mean": var_data["gauss_psfReconvolved_T"],
+            "flags": var_data["gauss_shape_flags"], # TO BE ADDRESSED!
```
joezuntz: Can you explain the comment more?

txpipe/ingest/metaDetect.py
```diff
@@ -0,0 +1,324 @@
...
+
+        chunk_size = self.config("chunk_rows")
+        pf = pq.ParquetFile(file_path)
+        for batch in pf.iter_batches(columns=input_columns, batch_size=chunk_size):
+            shear_data = process_metadetect_data(batch)
+            
+            if not created_files:
+                created_files = True
+                variants = {
+                    "ns": len(shear_data["ns"]),
```
joezuntz: Let's leave this as "ns" and change the rest of the code before merging since we have to do it anyway.

txpipe/ingest/metaDetect.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def write_output(self, outfile, group, data):
+        g = outfile[group]
+        for variant in ["ns", "1p", "1m", "2p", "2m"]:
+            k = g[variant]
+            for name, col in data[variant].items():
+                # replace masked values with nans
+                if np.ma.isMaskedArray(col):
+                    col = col.filled(np.nan)
+                k[name].append(col)  #NOT SURE THIS WORKS EITHER TBD
```
joezuntz: Have you checked this as the comment says? Delete comment if so.

txpipe/ingest/MetaDetect.py
```diff
@@ -0,0 +1,324 @@
...
+        for variant in ["ns", "1p", "1m", "2p", "2m"]:
+            k = g[variant]
+            for name, col in data[variant].items():
+                # replace masked values with nans
+                if np.ma.isMaskedArray(col):
+                    col = col.filled(np.nan)
+                k[name].append(col)  #NOT SURE THIS WORKS EITHER TBD
+
+
+def sanitize(data):
```
joezuntz: Repeated from other file. Put in a separate utils file.

joezuntz (changes requested):

txpipe/ingest/lsst.py
empEvil: Sorry what do you mean?

txpipe/ingest/MetaDetect.py
empEvil: It is meant for DP2, but can add that to it (it should work on DP1 as well because they look the same)

txpipe/ingest/lsst.py
empEvil: Right, I wasn't sure if that was the only flag we needed, or if we need to add a bunch of the flags instead.

txpipe/ingest/MetaDetect.py
empEvil: it exists in `lsst.py` so will import from there

txpipe/ingest/lsst.py
joezuntz: You wrote that it might not be needed, and I think that's right because this whole data set should have the same metaStep value because that's how they are selected. So no need to save it.

txpipe/ingest/metaDetect.py
empEvil: How do I/we update the sample catalog that the CI uses to reflect this change?

joezuntz: Yes, will need to do a whole replacement process. A bit fiddly.

empEvil:
@joezuntz I believe it is updated to work end to end

myamamoto26:
Thanks for the work on this! Some immediate comments and questions I thought about are, 

For ingest/metaDetect.py, 
- Would keeping 'ns' as a noshear catalog cause troubles downstream? Everything in the downstream currently uses '00'. 
- We might want to implement cuts based on flags during ingestion stage to keep the ingested catalogs light. It's fine for DP2 but would be significantly more objects for DR1. I don't think we want to do anything with objects with flags!=0, so I think that's okay to do.

empEvil:
I already went ahead and changed the down stream to 'ns'. 
hmm the ingestion is thought of something we do only once and then people will just use the centrally ingested catalog. But for DR1 we can do some kind of selection if we want

myamamoto26:
> I already went ahead and changed the down stream to 'ns'. 

OK! 

> hmm the ingestion is thought of something we do only once and then people will just use the centrally ingested catalog. But for DR1 we can do some kind of selection if we want

Right so wouldn't it be useful to have flags==0 for all objects in ingested catalogs? Or are there cases where we want to select objects based on different flags, in that case I agree we want to leave this in the selector stage.

myamamoto26:
> https://github.com/LSSTDESC/TXPipe/blob/MetaDetect_ingest/txpipe/ingest/metaDetect.py#L225

Is it intentional to leave out image_flags, bmask_flags and ormask_flags?

empEvil:
not intentional no. But it leads into the one unresolved question, of how do we want flags look for this?

myamamoto26:
Do you mean which flags to select on and which value?

empEvil:
yes, currently we have a `flags` that is used downstream and it is currently set to ` "flags": var_data["gauss_shape_flags"], `. But it is also noted that we might want something different

myamamoto26:
> yes, currently we have a flags that is used downstream and it is currently set to "flags": var_data["gauss_shape_flags"],. But it is also noted that we might want something different

Yeah we do want to use a set of flags (I think all the flags from cell-based coadd and ngmix steps), and so my suggestion was to select objects with *flags=0 during the ingestor stage and set `flags=0 ` for the downstream catalog.

joezuntz:
I just fixed the other issues in master so if you merge it into here it should hopefully start to work.

myamamoto26:
For the flags, I still think it's relevant which flags to output in the ingested catalog before we merge this PR into main, because the ingested catalog right now doesn't have all the information for us to select objects in the next TXPipe steps.

mirarenee:
What I have available for the DP1 metadetect should now also include fluxes in z (in case that was originally discluded due to the limited z coverage in the A360 field)

txpipe/diagnostics.py
```diff
@@ -712,16 +736,16 @@ def plot_g_histogram(self):
             if data is None:
                 break
 
-            qual_cut = data["bin"] != -1
+            #qual_cut = data["bin"] != -1
```
joezuntz, line 727: Do we not need a qual_cut of some kind here?

txpipe/ingest/lsst.py
```diff
@@ -64,3 +64,85 @@ def process_shear_data(data):
...
+
+def sanitize(data):
+    """
+    Convert unicode arrays into types that h5py can save
+    """
+    # convert unicode to strings
+    if data.dtype.kind == "U":
+        data = data.astype("S")
+    # convert dates to integers
+    elif data.dtype.kind == "M":
```
joezuntz, line 139: It's probably best to convert the dates to opaque types to avoid losing information - converting it to integers apparently depends on some underlying storage details.

https://docs.h5py.org/en/stable/special.html#storing-other-types-as-opaque-data

txpipe/ingest/metaDetect.py
```diff
@@ -0,0 +1,317 @@
+from ..base_stage import PipelineStage
```
joezuntz, line 1: I think a lot of this file is copied from dp1.py? If so shared things should be moved to another file and imported by each file, rather than duplicated.

txpipe/ingest/metaDetect.py
```diff
@@ -0,0 +1,317 @@
...
+        if self.config["select_field"]:
+            tracts = DP1_TRACTS[self.config["select_field"]]
+        elif self.config["select_tracts"]:
+            tracts = self.config["select_tracts"]
+        elif self.config["cosmology_tracts_only"]:
+            tracts = DP1_COSMOLOGY_TRACTS
+        else:
+            tracts = ALL_TRACTS
+
+        #n = self.get_catalog_size(butler, "ShearObject")
```
joezuntz: remove old commented-out code.

txpipe/ingest/metaDetect.py
```diff
@@ -0,0 +1,317 @@
...
+                splitter = MetaDetectSplitter(group, columns, variants, dtypes=dtypes)
+
+            for variant in ["ns", "1p", "1m", "2p", "2m"]:
+                splitter.write_bin(shear_data[variant], variant)
+            print(f"Processing chunk {i + 1} / {n_chunks}")
+
+        splitter.finish()
+        shear_outfile.close()
+
+    def file_run(self):
```
joezuntz: This is very very similar to the butler_run function above. You can make it one function and decide in an `iterate_data` method (or whatever) how the chunks should be read. Everything after that is the same. AI can fix that really well if you want.

txpipe/ingest/metaDetect.py
```diff
@@ -0,0 +1,317 @@
...
+            #"pgauss_T_flags",
+            "gauss_flags",
+            "pgauss_flags",
+            "gauss_shape_flags",
+            "is_primary"
+ 
+        ]
+        return input_columns
+
+    def setup_output(self, tag, group, first_chunk):
```
joezuntz: It looks like setup_output and write_output aren't used.

txpipe/ingest/metaDetect.py
```diff
@@ -0,0 +1,317 @@
...
+            k = g[variant]
+            for name, col in data[variant].items():
+                # replace masked values with nans
+                if np.ma.isMaskedArray(col):
+                    col = col.filled(np.nan)
+                k[name].append(col)
+
+
+
+# Outstanding issues!
```
joezuntz: Are these still outstanding?

txpipe/shear_calibration/names.py
```diff
@@ -1,4 +1,4 @@
-META_VARIANTS = ["00", "1p", "1m", "2p", "2m"]
+META_VARIANTS = ["ns", "1p", "1m", "2p", "2m"]
```
joezuntz, line 1: We should use this throughout the code.

txpipe/diagnostics.py
```diff
@@ -329,7 +332,13 @@ def plot_psf_shear(self):
 
         # Include a small shift to be able to see the g1 / g2 points on the plot
         dx = 0.1 * (psf_g_edges[1] - psf_g_edges[0])
-        idx = np.where(np.isfinite(mu1))[0]
+        idx = np.where(
```
joezuntz: We probably want a function to do this for here and below. Something like where_all_finite(arg, *args).

joezuntz (changes requested):

joezuntz:
Done my review. I'll ask the robot to do one too.

txpipe/shear_calibration/names.py
empEvil: you mean as we should be importing the naming from one central file every time we do anything with the variants?

txpipe/ingest/metaDetect.py
empEvil: Should be fixed, I am considering if we should drop the file run completely. I don't think it will be used

txpipe/ingest/metaDetect.py
empEvil: I think you are right, we can remove them

txpipe/ingest/lsst.py
empEvil: Sounds good, will change it

txpipe/diagnostics.py
empEvil: we might, but I don't actually think this is accessed anywhere else in the code... So I think I commented it out because it was causing issues, and wasn't being used

txpipe/diagnostics.py
empEvil: Not quite sure what you mean? are you wanting to move the this into a more general support function?

joezuntz: Repeatedly in this file we do `idx = np.where(isfinite(a) & isfinite(b) & isfinite(c) ...)` so we probably want:
```
def where_all_finite(array, *more_arrays):
    finite = np.isfinite(array)
    for arr in more_arrays:
        finite &= np.isfinite(arr)
    return np.where(finite)[0]
```

then we can just do 

```
idx = where_all_finite(mu1, mu2, std1, std2)
```
and similar.

txpipe/shear_calibration/names.py
joezuntz: Yep!

txpipe/ingest/lsst.py
```diff
@@ -64,3 +65,88 @@ def process_shear_data(data):
     output["psf_g2"] = psf_g2
 
     return output
+
+
+def process_metadetect_data(data, flag_list, flag_exclusion):
+    output = {}
+    for variant in META_VARIANTS:
+        var_data = data[data["metaStep"] == variant]
+        var_data = sanitize(var_data)
```
joezuntz, line 74: It looks like `sanitize` below is designed to operate on a single array whereas here it is sent either a dict of arrays or a structured array, and in either case won't work.

txpipe/ingest/lsst.py
```diff
@@ -64,3 +65,88 @@ def process_shear_data(data):
...
+
+        var_output = {
+            "ra": var_data["ra"],
+            "dec": var_data["dec"],
+            "id": var_data["shearObjectId"],
+            "object_mask_fraction": var_data["mfrac"],
+            #"n_epoch": var_data["nEpochCell"],
+            "g1": var_data["gauss_g1"],
+            "g2": var_data["gauss_g2"],
+            "g1_err": var_data["gauss_g1_g1_Cov"],
```
joezuntz: This will be the variance whereas g1_err is expected to be the standard deviation.

txpipe/ingest/metaDetect.py
```diff
@@ -0,0 +1,185 @@
...
+        "cosmology_tracts_only": StageParameter(bool, True, msg="Use only cosmology tracts."),
+        "select_field": StageParameter(str, "", msg="Field to select (overrides cosmology_tracts_only)."),
+        "select_tracts": StageParameter(list, [], msg="list of tracts (overrides cosmology_tracts_only, but not select_field)."),
+        "collections": StageParameter(str, "LSSTComCam/DP1", msg="Butler collections to use."),
+        "exclusion_flag": StageParameter(bool, False, msg="Decide if flags are used for exclusion or just flagged."),
+        "flag_list": StageParameter(list, ["is_primary"], msg="list of flags to use for combined.")
+        }
+
+    def run(self):
+        if self.config["use_butler"]:
```
joezuntz: If this is the only option then you can just delete the config option and rename butler_run to run.

joezuntz (changes requested):

txpipe/ingest/lsst.py
empEvil: good point. Do we want to save the covariance or should we just take the square root?

joezuntz: I don't think we are currently using either directly (only through the weights). So for simplicity you can just save the three covariance components (since it's symmetric there's no point saving 12 and 21)

empEvil: okay, so we should probably just rename them to `g1_var` and `g2_var` instead of err

joezuntz (approved):

PR #478, opened 2026-07-02
Fix cosmoDC2 ingestion

eduardojsbarroso:
In this pull request there are two things being changed:

- Adding true redshift information to all metadetect variants
- Adding i-band magnitude cuts.

### DIFF
--- a/txpipe/ingest/mocks.py
+++ b/txpipe/ingest/mocks.py
@@ -41,6 +41,8 @@ class TXCosmoDC2Mock(PipelineStage):
         ),
         "apply_mag_cut": StageParameter(bool, False, msg="Apply magnitude cut for descqa comparison."),
         "Mag_r_limit": StageParameter(float, -19, msg="Magnitude r limit for object selection."),
+        "apply_i_mag_cut": StageParameter(bool, False, msg="Apply observed i-band magnitude cut."),
+        "Mag_i_limit": StageParameter(float, 24.25, msg="Observed i-band magnitude limit for object selection."),
         "metadetect": StageParameter(
             bool, True, msg="Whether to make a metadetect-style catalog (True) or metacal (False)."
         ),
@@ -158,10 +160,21 @@ class TXCosmoDC2Mock(PipelineStage):
             # metacal info, saving us some time simulating un-needed objects
             if self.config["snr_limit"] > 0:  # otherwise there is no need to run this function which is slow
                 self.remove_undetected(data, mock_photometry)
+            # "apply_mag_cut" and "apply_i_mag_cut" should not be combined
+            # as one applies cut the data and the other both data and photometry
+            if self.config["apply_i_mag_cut"]:
+                sel = mock_photometry["mag_i"] < self.config["Mag_i_limit"]
+                ndet = sel.sum()
+                ntot = sel.size
+                fract = ndet * 100.0 / ntot
+                print(f"{ndet} objects pass observed i-band magnitude cut out of {ntot} objects ({fract:.1f}%)")
 
+                for key in list(mock_photometry.keys()):
+                    mock_photometry[key] = mock_photometry[key][sel]
+                for key in list(data.keys()):
+                    data[key] = data[key][sel]
             if self.config["apply_mag_cut"]:
                 self.apply_magnitude_cut(data)
-
             if self.config["metadetect"]:
                 mock_shear = self.make_mock_metadetect(data, mock_photometry)
             else:
@@ -273,14 +286,12 @@ class TXCosmoDC2Mock(PipelineStage):
             "dec",
             "psf_g1",
             "psf_g2",
-            "psf_g1",
-            "psf_g2",
             "psf_T_mean",
             "weight",
         ) + band_variants("riz", "mag", "mag_err", shear_catalog_type="metadetect")
 
-        # Store the truth values only for the primary catalog
-        cols += ["00/true_g1", "00/true_g2", "00/redshift_true"]
+        # Store the truth shear only for the primary catalog, true redshift for all variants
+        cols += ["00/true_g1", "00/true_g2"] + metadetect_variants("redshift_true")
 
         # Make group for all the photometry
         group = metacal_file.create_group("shear")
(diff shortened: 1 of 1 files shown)


txpipe/ingest/mocks.py
```diff
@@ -273,14 +286,12 @@ def setup_metadetect_output(self, metacal_file, target_size):
             "dec",
             "psf_g1",
             "psf_g2",
-            "psf_g1",
-            "psf_g2",
```
eduardojsbarroso, line 277: This was duplicated

txpipe/ingest/mocks.py
```diff
@@ -306,8 +317,6 @@ def setup_metacal_output(self, metacal_file, target_size):
                 "dec",
                 "psf_g1",
                 "psf_g2",
-                "psf_g1",
-                "psf_g2",
```
eduardojsbarroso, line 310: This was duplicated

txpipe/ingest/mocks.py
```diff
@@ -619,16 +632,6 @@ def make_mock_metadetect(self, data, photo):
...
-            "00/psf_g1": zero,
-            "1p/psf_g1": zero,
-            "1m/psf_g1": zero,
-            "2p/psf_g1": zero,
-            "2m/psf_g1": zero,
-            "00/psf_g2": zero,
-            "1p/psf_g2": zero,
-            "1m/psf_g2": zero,
-            "2p/psf_g2": zero,
-            "2m/psf_g2": zero,
```
eduardojsbarroso, line 631: This was duplicated

txpipe/ingest/mocks.py
```diff
@@ -1063,7 +1068,8 @@ def make_mock_photometry(n_visit, bands, data, unit_response):
     output["dec"] = data["dec"]
     output["id"] = data["galaxy_id"]
     output["extendedness"] = np.ones(nobj)
-
+    if "redshift_true" in data:
+        output["redshift_true"] = data["redshift_true"]
```
eduardojsbarroso, line 1071: Before, the code already used `photo["redshift_true"]`, but this information was being only passed as a column to `data["redshift_true"]`, so I added this line here to make sure the mock photometry also has the redshift information

txpipe/ingest/mocks.py
```diff
@@ -158,10 +160,21 @@ def run(self):
             # metacal info, saving us some time simulating un-needed objects
             if self.config["snr_limit"] > 0:  # otherwise there is no need to run this function which is slow
                 self.remove_undetected(data, mock_photometry)
-
+            # "apply_mag_cut" and "apply_i_mag_cut" should not be combined
+            # as one applies cut the data and the other both data and photometry
```
eduardojsbarroso: I did this separate i-band cut. I do not understand why the `apply_magnitude_cut` function does only a cut in data and not in the photometry as well, but it seems to be a absolute magnitude cut and not observed one, like the one I want here.

I know other stages, such as `TXRandomForestLensSelector` [here](https://github.com/LSSTDESC/TXPipe/blob/e47a01ff54f2dd30eb7365f87fe10a4dc11af604/examples/cosmodc2/config.yml#L165) has other observed magnitude cuts, but we do not have this for clusters and I wanted to have a magnitude cut in the ingestion phase of cosmoDC2 if that is ok.

eduardojsbarroso:
Hi @joezuntz , I found some other problems in this code and I marked everything in the comments. Let me know if this PR is ok or if the magnitude cut should strictly be added to [CLuster Source Selection](https://github.com/LSSTDESC/TXPipe/blob/e47a01ff54f2dd30eb7365f87fe10a4dc11af604/txpipe/extensions/cluster_counts/sources_select_compute.py).

My guess is that it is ok to have this cut here since it is only cosmoDC2 and we can think about implementing this in the cluster extensions later for dp1-dp2.

txpipe/ingest/mocks.py
```diff
@@ -41,6 +41,8 @@ class TXCosmoDC2Mock(PipelineStage):
         ),
         "apply_mag_cut": StageParameter(bool, False, msg="Apply magnitude cut for descqa comparison."),
         "Mag_r_limit": StageParameter(float, -19, msg="Magnitude r limit for object selection."),
+        "apply_i_mag_cut": StageParameter(bool, False, msg="Apply observed i-band magnitude cut."),
+        "Mag_i_limit": StageParameter(float, 24.25, msg="Observed i-band magnitude limit for object selection."),
```
joezuntz: Could you make the default very large so that there is no change in the default behaviour?

txpipe/ingest/mocks.py
joezuntz: No problem, it's fine to include, though note that this is just the ingestion stage right at the start where you typically want to ingest anything that you might need later. Then cuts to actual science samples usually happen later.

txpipe/ingest/mocks.py
```diff
@@ -158,10 +160,21 @@ def run(self):
             # metacal info, saving us some time simulating un-needed objects
             if self.config["snr_limit"] > 0:  # otherwise there is no need to run this function which is slow
                 self.remove_undetected(data, mock_photometry)
-
+            # "apply_mag_cut" and "apply_i_mag_cut" should not be combined
+            # as one applies cut the data and the other both data and photometry
+            if self.config["apply_i_mag_cut"]:
```
joezuntz: If you want this not to trigger then you could also set the default to zero.

joezuntz (changes requested):
Looks good - just one quick request.

eduardojsbarroso:
@joezuntz  Hey joe, actually I thought about it and for the work I have with true shapes ann true shear, I needed a true i band cut and not the photometry one to be consistent. Also then it makes more sense to make the cut here, since the mag i from the cosmodc2 catalog is never saved with this ingestion. So it works just like the mag r absolute cut we have.

Also, I changed then this abolute cut to be before the photometry. How it was before, we made the photometry for all the catalog objects, and then only cut the data column, and it gives a mismatch on the size of the data vs photometry, which was breaking the data iterator later.

The default is still `False` to apply this cut, and I set a magnitude cut that will change nothing if the user does not change the value to positive ones.

joezuntz (approved):

joezuntz:
Looks good - ready for me to merge?

eduardojsbarroso:
> Looks good - ready for me to merge?

Let's go :)

PR #480, opened 2026-07-09
basic selections in base.py, mdet specific selections in metadetect.py

myamamoto26:
- added a few more basic selections that are common between different catalogs (magnitude and color cuts)
- added metadetect specific initial selections to metadetect.py, which wraps the original selection function.

### DIFF
--- a/txpipe/source_selection/base.py
+++ b/txpipe/source_selection/base.py
@@ -335,13 +335,6 @@ def select_weak_lensing_sample(data, config, calling_from_select=False):
     # output is different in each case
     s2n_cut = config["s2n_cut"]
     T_cut = config["T_cut"]
-    mag_g_cut = config["mag_g_cut"]
-    mag_r_cut = config["mag_r_cut"]
-    mag_i_cut = config["mag_i_cut"]
-    mag_z_cut = config["mag_z_cut"]
-    gmr_cut = config["g-r_cut"]
-    rmi_cut = config["r-i_cut"]
-    imz_cut = config["i-z_cut"]
 
     verbose = config["verbose"]
     variant = data.suffix
@@ -366,20 +359,10 @@ def select_weak_lensing_sample(data, config, calling_from_select=False):
     sel &= s2n > s2n_cut
     f3 = sel.sum() / n0
 
-    # We should also have some crazy color cuts and magnitude cuts which should come from PZ group
-    sel &= (data["mag_g"] < mag_g_cut) & \
-        (data["mag_r"] < mag_r_cut) & \
-        (data["mag_i"] < mag_i_cut) & \
-        (data["mag_z"] < mag_z_cut) & \
-        (np.abs(data["mag_g"] - data["mag_r"]) < gmr_cut) & \
-        (np.abs(data["mag_r"] - data["mag_i"]) < rmi_cut) & \
-        (np.abs(data["mag_i"] - data["mag_z"]) < imz_cut)
-    f4 = sel.sum() / n0
-
     # Finally we want objects that have been put into any of our other
     # tomographic bins
     sel &= data["zbin"] >= 0
-    f5 = sel.sum() / n0
+    f4 = sel.sum() / n0
 
     # Print out a message.  If we are selecting a 2D sample
     # this is the complete message.  Otherwise if we are about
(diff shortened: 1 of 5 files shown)


joezuntz:
I think this can be re-targeted at the master branch now?

joezuntz:
Thanks for these!

I think it would be better to have a specific DP2 selector as a subclass of MetadetectSelector in its own new file, with the new content in that. Then we can iterate on that specifically as the data comes in and we find out what new selections we need.

myamamoto26:
@joezuntz Are you thinking of something like dp2.py where we can make both Metadetect and AnaCal DP2 selections?

joezuntz:
Yep - exactly.

txpipe/source_selection/base.py
```diff
@@ -55,6 +55,13 @@ class TXSourceSelectorBase(PipelineStage):
             required=True,
             msg="Signal-to-noise cut threshold for object selection",
         ),
+        "mag_g_cut": StageParameter(float, required=True, msg="Magnitude cut threshold for object selection"),
+        "mag_r_cut": StageParameter(float, required=True, msg="Magnitude cut threshold for object selection"),
+        "mag_i_cut": StageParameter(float, required=True, msg="Magnitude cut threshold for object selection"),
+        "mag_z_cut": StageParameter(float, required=True, msg="Magnitude cut threshold for object selection"),
+        "g-r_cut": StageParameter(float, required=True, msg="Color cut threshold for object selection"),
+        "r-i_cut": StageParameter(float, required=True, msg="Color cut threshold for object selection"),
+        "i-z_cut": StageParameter(float, required=True, msg="Color cut threshold for object selection"),
```
empEvil: I don't think these cuts should be required

joezuntz: If we do want them then they could go in the new dp2.py file

myamamoto26: OK i'll keep the magnitude and color cuts in dp2.py then.

myamamoto26: But it does feel a bit odd that the size ratio selection is required and magnitude/color cuts are not. These mag/color cuts are common and will be required by both metadetect and anacal catalogs.

myamamoto26: I am going to put mag/color cuts in dp2.py for now so that we can merge this PR into main, before the sprints.

myamamoto26: @joezuntz @empEvil Updated now!

myamamoto26:
Is it normal to fail these CIs?

txpipe/source_selection/base.py
joezuntz: It looks like these ones are still in base.py - could you put them just in dp2.py?

joezuntz:
The error is: `ValueError: Missing configuration option mag_g_cut` because the new options are defined on all the source selectors because of the inheritance - if you move the new options to the new class it should fix it. You can also inherit the parent ones with something like:
```
config_options = TXSourceSelectorBase.config_options | {
# new options
}
```

txpipe/source_selection/base.py
myamamoto26: sorry, I forgot to remove the config options bit.

joezuntz:
The remaining error is unconnected with these changes. I’ve tried re-running to fix it as that sometimes works. I haven’t been able to track the cause apart from it being MPI-related.

myamamoto26:
CIs passed now! 
I think it was failing because the hdf5 catalog used for tests didnt have the columns needed. I added data_iterator function in dp2.py to add more columns.

joezuntz:
Looks good - just getting a quick AI review.

txpipe/source_selection/dp2.py
```diff
@@ -0,0 +1,138 @@
+from .metadetect import TXSourceSelectorMetadetect
+from .base import select_weak_lensing_sample, TXSourceSelectorBase
+from ..shear_calibration import MetaDetectCalculator
+from ceci.config import StageParameter
```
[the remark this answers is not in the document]
myamamoto26: I changed this in the latest push.

joezuntz:
I've dealt with the remaining issues copilot raise. The only major point is that I've changed `g-r_cut` etc to `gr_cut`. I'm happy to merge if you are!

joezuntz (approved):

PR #485, opened 2026-07-20
Update mapping organization

joezuntz:
Updates mapping stages in two ways:
- make the lens maps directly from the binned_lens file rather than the original catalogs
- split up the aux lensing mapping into multiple stages


txpipe/masks.py
empEvil, line 307: Should this be called Custom, when it specifically requires a DES footprint as an input?

PR #492, opened 2026-07-25
Adding TXMeanShearSurveyProperties (mean shear vs. survey property maps)

ShrihanSolo:
### Summary

Adds a new null-test stage, `TXMeanShearSurveyProperties`, that measures the mean calibrated shear as a function of survey-property maps (provided PSF properties, depth maps, sky background, etc.). This is analogous to the existing shear-vs-PSF / tomographic null tests but generalized to arbitrary survey-property maps.

### Method

- Binning: For each property map, only source galaxies within the map footprint are used. `bin_edges` is defined based on the survey property values of actual galaxies in the shear catalog. `outlier_fraction` trims (by default, 5%) outlier survey properties to determine the outer bin edges, and the remaining `nbins` divide this range evenly.
- Mean shear: Existing `MeanShearInBins` machinery is used. This means, whether metacal, metadetect, or lensfit/HSC, response R or m is recomputed per property bin from the sheared columns to include the effects of selection.
- A straight-line fit (slope ± error, χ²/dof) is reported per property for both g1 and g2.
- Two-panel per-property figure generated: the property map visualized alongside the mean-shear trend.

### Consistency with existing TXPipe stages

I strived for consistency with the pipeline, and so reused TXPipe's existing shear-diagnostic machinery as much as possible:

- Binned mean shear is computed with `MeanShearInBins` (txpipe/shear_calibration/mean_shear_in_bins.py) — the same calibrator that backs `TXSourceDiagnosticPlots` and `TXLensDiagnosticPlots` (txpipe/diagnostics.py). These stages already bin mean shear against PSF size/ellipticity, SNR, size, etc.  This stage swaps these for a survey-property map value (for any survey property provided). Reusing it also means metacal/metadetect/lensfit/HSC calibration is handled identically to the established null tests.
- Metacal response is recomputed per bin via `MetacalCalculator` (txpipe/shear_calibration/calibration_calculators.py), consistent with source selection (txpipe/source_selection/metacal.py) and the diagnostic stages mentioned already.
- Metadetect per-variant handling follows `TXSourceDiagnosticPlots`: ra/dec, shear, and the selection/bin column are read per sheared variant, using `metadetect_variants` / `META_VARIANTS` from txpipe/shear_calibration.
- Plotting conventions match the diagnostics code: `TXSourceDiagnosticPlots`, and finite-value masking reuses `where_all_finite` from txpipe/diagnostics.py.
- Uses `combined_iterators`, `parallel = True` MPI reduction with chunking, `FileCollection` for per-property plots + an HDF5 summary,  follows the same patterns as the existing diagnostic/null-test stages.

### Inputs / outputs

- Inputs: shear_catalog, shear_tomography_catalog, aux_source_maps, aux_lens_maps. By default, the code runs on the aux-maps outputs (psf, flags, depth) at a minimum, even with no additional survey properties provided. However, if not desired, these inputs are also optional — alias either/both to "none" to skip, so the test can run on external maps alone. E.g. in pipeline.yml:

```
  - name: TXMeanShearSurveyProperties
    aliases:
        aux_lens_maps : none
        aux_source_maps : none
```
where none is defined with none: none later in the file (as in most of the example pipeline.yml)

- Outputs: `shear_survey_property_null.hdf5` (HDF5: per-property means, errors, slopes, χ²) and `shear_survey_property_plots/` directory (per-property PNG diagnostics).

### Config options

`nbins` (20), `chunk_rows` (100,000), `delta_gamma` (for response, 0.02), `outlier_fraction` (5% for each tail), `external_maps_dir` (path to .hs healsparse files to test), `map_plot_nside` (map-panel resolution for plotting only).

### Testing

Adds txpipe/test/test_survey_property_null_tests.py covering the stage.

### Other notes

- Added stage in txpipe/__init__.py.
- Generative AI was used to help with this PR: specifically for using consistency of conventions/not duplicating effort, generating docstrings/comments, the plotting code, and the testing code.

### Example results on DP1 Example

Pipeline was successfully run on `dp1_example` [code](https://github.com/LSSTDESC/SprintDays2026/issues/8) and DP1 healsparse survey property maps:

image
image
image


ShrihanSolo:
Possible future update plan: Create an alternative to MeanShearInBins that uses jackknife to estimate uncertainty on mean by removing patches of the footprint.

ShrihanSolo:
Since the previous PR, some changes were made to `aux_source` and `aux_lens`, prompting an update to the PR. I have made the following changes.

**Inputs Updated**

```
("survey_property_maps", FileCollection),   # from TXIngestDataPreview1
("psf_maps", MapsFile),                     # from TXPSFMaps
("depth_map", MapsFile),                    # from TXDepthMaps
```
Now, the inputs are rewired to take `TXPSFMaps` and `TXDepthMaps` maps instead of `aux_source` and `aux_lens`. These inputs are optional, and each can be removed as follows if not needed in pipeline.yml:

```
 inputs:
        survey_property_maps : none
        psf_maps : none
        depth_map: none
```

**Tags Now Used**

The earlier `external_maps_dir` has been made into a fallback parameter, and the main path used for survey property maps is provided in the `pipeline.yml` inputs section with the `survey_property_maps` tag, rather than in the config file. 

**Bands Config Added**

Additionally, a new "bands" config was added, since there may be many survey property maps, and an analysis may only be need to run on, say i-band. The input here is a list of bands in the config.py file.

```
TXMeanShearSurveyProperties:
    bands: [i]
```

The current commit was tested to run on DP1 on NERSC successfully, with outputs, e.g.

image

Titles have also been shortened for plotting and filenames if they follow the standard format.

PR #494, opened 2026-08-05
Fixed the subtle bug in the quantilization

empEvil:
We noticed that on the DP2 diagnostics, all the PSF quantiles where positive, while we would expect atleast the one to be negative. When comparing running with and without DASK it was noted there was a difference. This PR should fix that issue

### DIFF
--- a/txpipe/diagnostics.py
+++ b/txpipe/diagnostics.py
@@ -57,13 +57,22 @@ class TXDiagnosticQuantiles(PipelineStage):
         with self.open_input("shear_catalog", wrapper=True) as f:
             group = f.get_primary_catalog_group()
 
+        cat_type = read_shear_catalog_type(self)
         # We canonicalise the names here
-        col_names = {
-            "psf_g1": f"{group}/psf_g1",
-            "psf_T_mean": f"{group}/psf_T_mean",
-            "s2n": f"{group}/s2n",
-            "T": f"{group}/T",
-        }
+        if cat_type == "metadetect":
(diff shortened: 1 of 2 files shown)


joezuntz:
This is failing because the test data doesn't have psf_original. You could add the prefix/suffix as a config option maybe?

empEvil:
So now it failed on:
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/runner/work/TXPipe/TXPipe/txpipe/__main__.py", line 6, in <module>
    PipelineStage.main()
  File "/home/runner/work/TXPipe/TXPipe/conda/lib/python3.12/site-packages/ceci/stage.py", line 526, in main
    stage.execute(args)
  File "/home/runner/work/TXPipe/TXPipe/conda/lib/python3.12/site-packages/ceci/stage.py", line 699, in execute
    stage.run()
  File "/home/runner/work/TXPipe/TXPipe/txpipe/diagnostics.py", line 105, in run
    masked = masked.compute_chunk_sizes()
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'MockDaskArray' object has no attribute 'compute_chunk_sizes'

But only for one of the runs?

joezuntz:
You'll need to add the `compute_chunk_sizes` method to the `MockDaskArray` class, just returning self.

joezuntz (approved):

PR #501, opened 2026-09-22
Use skyproj for map plots

joezuntz:
Updates plotting to use the much nicer skyproj which plays cleanly with HealSparse maps.

Also tidies up the mapping code a little.


joezuntz:
The crash here seems to be to do with an illegal instruction with NaMaster, unrelated to the PR.

txpipe/data_types.py
```diff
@@ -267,7 +268,7 @@ def read_map(self, map_name):
         info = self.read_map_info(map_name)
         pixelization = info["pixelization"]
         if pixelization == "gnomonic":
-            m = self.read_gnomonic(map_name)
+            raise ValueError("Gnomonic maps are not currently supported.")
```
empEvil: Is there a reason why we are completely removing gnomonic?

txpipe/data_types.py
```diff
@@ -489,15 +490,12 @@ def write_map_pixval(self, map_name, pixel, value, metadata, nside_coverage=32):
             hdf5_group=f"maps/{map_name}/healsparse",
         )
 
-    def plot_healpix(
+    def plot(
```
empEvil: Why are we going from plot_healpix to plot? I assume it was named like that because we used to support multiple plot methods?

txpipe/map_plots.py
```diff
@@ -37,18 +34,12 @@ class TXMapPlots(PipelineStage):
...
         ("bright_object_map_plot", PNGFile),
     ]
     config_options = {
-        "projection": StageParameter(str, "cart", msg="Projection type for map plots (e.g., cart, moll)"),
-        "rot180": StageParameter(bool, False, msg="Whether to rotate the map by 180 degrees"),
-        "debug": StageParameter(bool, False, msg="Enable debug mode for plotting"),
-        "mask_threshold": StageParameter(
-            float, 0.0, msg="Threshold for masking pixels"
-        ),
+        "projection": StageParameter(str, "McBryde", msg="Skyproj projection type for map plots (e.g., McBryde, Mollweide)"),
```
empEvil: Is there a particular reason we are removing configuration options? while they might not be needed for DP2, couldn't there be other reasons to preserve them?

txpipe/map_plots.py
```diff
@@ -58,89 +54,31 @@ def run(self):
...
-                sys.stderr.write(f"Failed to make maps with method {m.__name__}")
-
-    def aux_source_plots(self):
-        """
-        Plot source auxiliary maps
+        self.source_plots()
+        self.lens_plots()
+        self.mask_plots()
+        self.aux_lens_plots()
+        self.psf_maps()
```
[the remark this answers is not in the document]
empEvil: Is there a reason to not isolate the failures?

txpipe/map_plots.py
```diff
@@ -58,89 +49,31 @@ def run(self):
...
         """
         import matplotlib.pyplot as plt
+        if self.get_input("psf_maps") == "none":
+            self.make_empty_plot("psf_map")
+            return
 
-        has_psf_maps = self.get_input("psf_maps") != "none"
-        has_flag_maps = self.get_input("flag_maps") != "none"
-
-        if has_flag_maps:
```
empEvil: Again, I am struggling to understand why we are removing options? won't some of these be needed by others?

txpipe/map_plots.py
```diff
@@ -252,43 +138,28 @@ def lens_plots(self):
...
                     plt.title(f"No map generated for {map_type}")
             return
 
         m = self.open_input("lens_maps", wrapper=True)
         rho = self.open_input("density_maps", wrapper=True)
         nbin_lens = m.file["maps"].attrs["nbin_lens"]
 
-        # If the maps require a degrade the reduction will be a weighted mean
-        # so we load the mask here at the same nside as the map (to be used as weights)
-        nside = rho.read_map_info("delta_0")["nside"]
```
empEvil: Will we not need degrading options?

txpipe/map_plots.py
```diff
@@ -348,18 +216,7 @@ def run(self):
...
-
-        # We don't want this to fail if some maps are missing.
-        for m in methods:
-            try:
-                m()
-            except:
-                if self.config["debug"]:
-                    raise
-                sys.stderr.write(f"Failed to make maps with method {m.__name__}")
+        self.aux_ssi_plots()
```
empEvil: Same question as before, we are going from isolating failures, to it all falling over?

empEvil (commented):
I am curious why we seem to be removing options, do we not want backwards compatible?

txpipe/map_plots.py
joezuntz: The rot180 thing was only needed because something was wrong with the old plotting when we straddled the ra=0 line, but skyproj deals with that internally.

We weren't using mask_threshold in practice.

txpipe/map_plots.py
joezuntz: See above!

txpipe/map_plots.py
joezuntz: No - that's the nice thing about skyproj - instead of plotting every pixel, which would be impossible at super high res, so you would need to downgrade. But skyproj instead looks up the sky pixel corresponding to each pixel in the output image, so it's faster and you can just pass in an nside=131027 map and it's fine!

txpipe/map_plots.py
joezuntz: As above!

PR #502, opened 2026-09-22
move imports inside twopoint_cluster.py

joezuntz:
The desc-stack environment doesn't have the full TXPipe environment and so imports can fail. This moves them to the functions where they are needed.


eduardojsbarroso (approved):
Looks good. This is actually how it suposed to be with Ceci right? So we can import different pipeline stages that will be run in different envs. This is some old code we have to check. Thank you Joe

joezuntz:
Thanks! Yes, that's right, that's currently the usual way to do things in ceci.

Python 3.15 introduces lazy imports, so maybe we can move to that at some point.

